PR #41, opened 2018-08-20
Update gettingstarted.md

John-A-Davies:
The relative directory of paxes from builds is `../paxes`, not `../../paxes`.


nannanli:
Thanks John! The update is now available on the docs site: https://zowe.github.io/docs-site/user-guide/gettingstarted.html

PR #42, opened 2018-08-20
Some peer edits

jhenryRocket:
Some peer edits
I think that the topic mvd-mvdandwindowmgt.md can be deleted as I don't see it being used in the ditamap.


nannanli:
@jhenryRocket  Agree, the mvd-mvdandwindowmgt.md is not used in ditamap or online doc. The one that is in use is mvd-desktopandwindowmgt.md. So it is ok to delete this topic.

PR #48, opened 2018-08-23
fixes #46

### LINKED ISSUE #46: Document the procedure to verify the Zowe package
Currently verifying the PAX file is a mandatory step and the procedure is documented only in the zowe.org right after users click "Agree" on the license page: https://projectgiza.org/Downloads/verify.html

Some users might assume that after saving the file, all is done on the download site. Then they go directly to read documentation without noticing that a verification step is required on the download site.  Therefore it makes sense to document the procedure in docs site User Guide: https://zowe.github.io/docs-site/user-guide/gettingstarted.html

Within this topic, further divide it into:
- Verifying the file (provide the current steps as well as link to download gpg)
- Transfer the file to z/OS

**Comments from Mike:** 
we should a) make it very obvious how to get from ‘download’ to ‘install docs’ (at the time when I did this, I had to hunt a little to find the docs) and b) the first step in the install docs should be to verify the download. We should state it as a step and I imagine some people will skip it - but that needs to be their choice not an accident.

jackjia-ibm:
Signed-off-by: Jack (T.) Jia <jack-tiefeng.jia@ibm.com>

### DIFF
--- a/docs/user-guide/gettingstarted.md
+++ b/docs/user-guide/gettingstarted.md
@@ -19,6 +19,7 @@ You can download the PAX file from the [Zowe website](https://zowe.org/download/
     **Notes:** 
     
     - The commands in the following steps are tested on both Mac OS X V10.13.6 and Ubuntu V16.04 and V17.10.
+    - Ensure that you have GPG installed. Click [here](https://www.gnupg.org/) to download and install GPG.
     - The `v.r.m` in the commands of this step is a variable. You must replace it with the actual PAX file version, for example, `0.9.0`.
 
     a. Verify the hash code.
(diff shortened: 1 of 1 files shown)


jackjia-ibm:
@nannanli good questions.

For link to gnupg.org, I agree. It may make the user easier to find the required tool.

For the whole download process, there are several issues we are trying to solve.

- we are trying to solve unfinished/broken download, so we have a new downloading tool for testing,
- I will suggest Alvin to add a simple files list break down by versions on the download website, so we can point to that page. Right now we don't have this link.
- The download pages in ProjectGiza.org is supposed to be temporary until we fix the close-source packages.

So I think in the next few months, we have to keep updating this section. But for now, https://projectgiza.org/Downloads/verify.html is the only place points to download the .asc, .sha512 files.

nannanli:
Thanks Jack for the clarification! The PR is merged now. We'll revisit this topic when the download process is updated.

PR #78, opened 2018-09-17
:memo: Corrections in install and profile creation

crshnburn:
* Change a Db2 reference to be CICS in the plug-in installation steps
* Add the necessary flags to the profile creation command.


JamesBauman:
@BrandonJenkins14 -

The corrections appear to be valid.  The cis (instead of cics) needs to be fixed in confluence too..

PR #170, opened 2018-10-19
fix prereqs

BrandonJenkins14:
Signed-off-by: BrandonJenkins14 <jenbr05@ca.com>


JamesBauman:
Note that we need to update `config.js` with the new topic title.....

JamesBauman:
If the content addresses the issue properly, please let me know. I'll copy as required and publish accordingly.

thx,

PR #229, opened 2018-11-08
fixed link again from other article

JamesBauman:
Signed-off-by: bauja03 <bauja03@ca.com>


JamesBauman:
corrected links to images
we moved the overview article so we need to remap the links to the images

JamesBauman:
locked and loaded, ready to go ... lets publish this content ....

PR #244, opened 2018-11-16
Update troubleshootinstall.md

John-A-Davies:


John-A-Davies:
There should be no dot in front of the word 'profile'.

PR #288, opened 2018-12-11
Added back doc about pre and post install scripts

nannanli:


stevenhorsman:
Joe has said he's not happy with the name post-install, so you should get his review.
These changes also cannot go in until 0.9.5, so I'm not sure if putting them in the What's new 0.9.4 is the best place.

nannanli:
If this is a 0.9.5 work item, we must hold on merging it to master until we publish 0.9.5. Also the What's new must go into 0.9.5 release notes (instead of the current 0.9.4).

Joe-Winchester:
I'd like us to hold off on delivering this until 0.9.5  Customers have been confused by the script names and also our instructions for when they should be executed.
Conceptually we have two scripts.  One in the /install folder that is not a pre-install.  It is a pre-req checker and it is optional  The second is in the /scripts folder and is a troubleshoot script, that reports any problems that may be causing zowe to be misfunctioning.  
@John-A-Davies - we can chat more about this when I hopefully get into the office later this week if that'd be helpful.  The big problem is that pre-install and how we described it led users to run it in 0.9.4 when it reported missing folders that weren't needed, and also it only worked with RACF.  Likewise post-install is not post install - it is trying to troubleshoot a runtime and can be run at any time so "install" being part of the name just confused customers

John-A-Davies:
@Joe-Winchester The pre- script works OK if you set the right INSTALL_DIR and the post- works as-is.  Not perfect but useful, although not customer-ready.  Leave them undocumented for now.

Developing and testing them to include ACF2 and Top Secret will take some time - hard to estimate that right now.

nannanli:
@Joe-Winchester @John-A-Davies Based on the disussion, it seems the two scripts can be called `zowe-prereq-verify.sh` and `zowe-install-troubleshoot.sh` respectively. Just my two cents. The descriptions should highlight that the steps are optional. I'd propose the following description:

Move the prereq checker script description to the **Prerequisite** section as first bullet and state clearly that it's only supprted for RACF users in this release. Or better move it early to the **System Requirement** section after users finish setting up various prereqs so they don't have to go back to check prereqs again before they start installation. 
> - **Optional:** You can check if your system meets the basic prerequisites for installing Zowe on your z/OS system if you use RACF. To do this, issue the following command with the current directory being the /install directory.
> `zowe-verify-pre-install.sh`
> The script writes its messages to your terminal window. The results are marked `OK`, `Info`, `Warning` or `Error`. Correct any reported errors and re-run the command before you run the `zowe-install.sh` script to install the Zowe runtime. The `zowe-verify-pre-install.sh`(tbd) script does not change any settings. You can run it as often as required before you install the Zowe runtime.

Then users would start installing Zowe runtime with `zowe-install.sh` script. Then about the troubleshoot script, we might want to clarify some points in our descriptions like: 
- Does it happen before users verify the independent components? 
- What aspects will it troubleshoot? 
- With this script, will users still have to verify each component? 

If its troubleshooting script, thinking that it will be a new section after installation but before Verifying installing as **`Troubleshooting installing Zowe runtime`**?

nannanli:
Hi @Joe-Winchester @John-A-Davies , I'm closing this PR as the doc updates will go to 0.9.5 and the script names will change. The new descriptions are now on the docs-staging branch and ready for your review. See this change: https://github.com/zowe/docs-site/commit/1d2af8c47b37eb53a125f960321c7e0ebd84a1c3 or the PDF file that's sent to you on Slack. Thanks!

PR #325, opened 2019-01-11
Fix broken code blocks and remove outdated diagram

janan07:


JamesBauman:
@janan07  publishing content updates....

PR #352, opened 2019-01-31
Create appendix

BrandonJenkins14:
I added Appendix to the nav. It contains everything we need from a legal perspective, including Bill of Materials and TPSRs for all components. 

- @jackjia-ibm I don't think it's an essential fix right away, but adding a nav element and moving canHideFirst attribute to the new element might have reintroduced our wrapping problem. 
- If anyone takes issue with having an Appendix, we can go another direction and add a link from the homepage to standalone pages that don't appear in any TOC. My vote is for Appendix though, who knows when we might need to add more content here. Also, we are close to GA so I don't want to add more last-minute work by re-doing this section if we are covered legally with what is here. 
- @MarkAckert @JamesBauman I removed the TOC from the tpsr.md file. It was failing broken link tests. Feel free to fix if you have any ideas.


docs/.vuepress/config.js
```diff
@@ -9,120 +9,138 @@ const PUBLISH_TARGET_PATH = (process.env.PUBLISH_TARGET_PATH || 'latest').replac
...
+      'user-guide/mvd-using.md',
+      'user-guide/api-mediation-api-catalog.md',
+      'user-guide/cli-usingcli.md',
+    ]
   },
   {
-    text: 'Contributing',
-    baseuri: '/contributing.html',
-    link: 'contributing.md',
-    canHideFirst: true,
```
jackjia-ibm: @BrandonJenkins14 you can keep the `canHideFirst: true,` on `Contributing` page. With this, both of Contributing and Appendix links will be hidden early than others.

PR #359, opened 2019-02-06
Update install-zos.md

John-A-Davies:
I've added a description of how to use the supplied `zowe-config-stc.sh` script.


nannanli:
Changed the base branch to docs-staging which hosts the 1.0 content.

PR #362, opened 2019-02-07
Fix springv1 guidelines

taban03:
Improve/fix spring onboarding guidelines.

### DIFF
--- a/docs/extend/extend-apiml/api-mediation-onboard-a-sprint-boot-rest-api-service.md
+++ b/docs/extend/extend-apiml/api-mediation-onboard-a-sprint-boot-rest-api-service.md
@@ -266,7 +266,7 @@ The first step to onboard a REST API with the Zowe ecosystem is to add enabler a
          * The service ID value must contain only lowercase alphanumeric characters.
          * The service ID cannot contain more than 40 characters.
          * The service ID is linked to security resources. Changes to the service ID require an update of security resources.
-         * The service ID must match spring.application.name.
+         * The service ID must match `spring.application.name` parameter.
 
          **Examples:**
 
--- a/docs/extend/extend-apiml/api-mediation-onboard-a-sprint-boot-rest-api-service.md
+++ b/docs/extend/extend-apiml/api-mediation-onboard-a-sprint-boot-rest-api-service.md
@@ -32,28 +32,11 @@ The Spring Boot API onboarding process follows these general steps. Further deta
 4. Test your changes.
 
 ## Add Zowe API enablers to your service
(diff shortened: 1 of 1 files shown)


docs/extend/extend-apiml/api-mediation-onboard-a-sprint-boot-rest-api-service.md
```diff
@@ -245,7 +245,8 @@ The first step to onboard a REST API with the Zowe ecosystem is to add enabler a
       spring:
           application:
               name: ${mfaas.discovery.serviceId}      
-
+    ```
+    **Note:** Add this configuration also to the `application.yml` used for testing.
```
janan07: Can you explain in this note why it is important to add this configuration to application.yml used for testing?

docs/extend/extend-apiml/api-mediation-onboard-a-sprint-boot-rest-api-service.md
```diff
@@ -256,15 +259,16 @@ The first step to onboard a REST API with the Zowe ecosystem is to add enabler a
...
 
         **Important!**  Ensure that the service ID is set properly with the following considerations:
 
-         * When two API services use the same service ID, the API gateway considers the services to be clones. An incoming API request can be routed to either of them.
+         * When two API services use the same service ID, the API Gateway considers the services to be clones. An incoming API request can be routed to either of them.
          * The same service ID should be set for only multiple API service instances for API scalability.
          * The service ID value must contain only lowercase alphanumeric characters.
          * The service ID cannot contain more than 40 characters.
          * The service ID is linked to security resources. Changes to the service ID require an update of security resources.
+         * The service ID must match the `spring.application.name` parameter.
```
JirkaAichler: Can we somehow highlight it? It's super important :-)

janan07: This is already a bullet under "Important". If it deserves more attention we can put this point at the top of the list and add an explanation of what happens if the ID does not match the spring.application.name parameter. Can you edit this point to explain what happens if the service ID and spring.application.parameter don't match?

taban03:
> Do we want change order of steps? Step order 2 -> 3 -> 1

Yes, good point, we should put the enabler configuration as last step

docs/extend/extend-apiml/api-mediation-onboard-a-sprint-boot-rest-api-service.md
janan07: If this is a necessary step, then it should be added as a step rather than a note.

PR #363, opened 2019-02-08
Minor corrections for installation of ZWESIS01 Server Security definitions.

hogstrom:


hogstrom:
bugger, wrong target.

PR #365, opened 2019-02-08
Update install-zos.md

JakubSkrivanek:
Troubleshooting for missing auto conversion. https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.3.0/com.ibm.zos.v2r3.bpxb200/setupa.htm

### DIFF
--- a/docs/user-guide/install-zos.md
+++ b/docs/user-guide/install-zos.md
@@ -505,6 +505,17 @@ By default, Zowe uses the runtime version that you most recently installed. To s
 /S ZOWESVR,SRVRPATH='$ZOWE_ROOT_DIR'
 ```
 
+When `ZOWESVR` procedure fails with `RC=0256` and `STDERR` contains:
+```
+.../scripts/internal/run-zowe.sh 1: FSUM7332 syntax error: got (, expecting Newline
+```
+BPX automatic code page conversion is not enabled. To enable auto conversion in `ZOWESVR` proc add the following into the `ZOWESTEP`:
+```
+//STDENV   DD *   
+_BPXK_AUTOCVT=ON  
+/*
+```
+
 To test whether the API Mediation Layer is active, open the URL: `https://<hostname>:7554`. 
 
 The port number 7554 is the default API Gateway port. You can overwrite this port in the `zowe-install.yaml` file before the `zowe-install.sh` script is run. See Step 2 in [Installing the Zowe runtime on z/OS](#installing-the-zowe-runtime-on-zos).


plavjanik:
Hello @JakubSkrivanek and @John-A-Davies!

I am looking into the [zowe-check-prereqs.sh](https://github.com/zowe/zowe-install-packaging/blob/42629bbba61954c70e7a1410ff4650e1c9c0c2d7/install/zowe-check-prereqs.sh#L538) script written by @John-A-Davies . This script is intended to be executed before the installation. It checks the value of `_BPXK_AUTOCVT` and if it is set then if issues a warning. 

My understanding is that Zowe installation expects the auto conversion to be off so all the files are created in the same encoding EBCDIC. Then the Zowe started tasks does not need `_BPXK_AUTOCVT=ON`. `_BPXK_AUTOCVT=ON` will help with the files that were stored in ASCII and tagged during installation but it can affect other files. Since the Zowe has been tested with without `_BPXK_AUTOCVT` then I would recommend to install Zowe without it.

@stevenhorsman, do you know if we support `_BPXK_AUTOCVT=ON` during installation or we expect to have it off?

@JakubSkrivanek, what has been the value of `_BPXK_AUTOCVT` during your installation? One way how to find out is to run `zowe-check-prereqs.sh` or issue `echo $_BPXK_AUTOCVT` in your z/OS UNIX environment.

JakubSkrivanek:
Hi, I have initially reported the issue for version 1.0.0 NB. My result of running the prereq scrip was:
Check the USS AUTOCVT                                                           
OK: OMVS AUTOCVT is set to OFF 
my  echo $_BPXK_AUTOCVT command showed value ON

I have moved to version 1.0.0 RC2 since. Running the prereq scrip now, I can see:
Check USS AUTOCVT
OK: OMVS AUTOCVT is set to OFF
Warning: _BPXK_AUTOCVT is set to ON
OK: _BPXK_CCSIDS is not set

I have changed the setting of  _BPXK_AUTOCVT  to stay on OFF which resulted in problems during run of the zowe install script.
FSUM7332 syntax error: got (, expecting Newline to be exact.

Looking at my environment set 
, I have path set including Rocket/python36.
There is SED available in that path which messes up the coding while running the installation script.
Commenting out all the paths to this folder and the BPXK, _CEE_RUNOPTS, and TAGs resulted into a correct run of the installation script.

John-A-Davies:
Thanks.  Please indicate which script produced the error `FSUM7332 syntax error: got (, expecting Newline`.  Please show me what that file looks like in ISPF 3.17 when you VIEW it and also when you type MF against its name.

The warnings from verify scripts about AUTOCVT are merely advisory - we don't want to change the user's environment permanently and we don't want to mandate or deprecate such settings.

plavjanik:
> FSUM7332 syntax error: got (, expecting Newline to be exact.
> 
> Looking at my environment set
> , I have path set including Rocket/python36.
> There is SED available in that path which messes up the coding while running the installation script.
> Commenting out all the paths to this folder and the BPXK, _CEE_RUNOPTS, and TAGs resulted into a correct run of the installation script.

I have reported the problems with `sed` in this item - https://github.com/zowe/zowe-install-packaging/issues/381. The workaround was to update the Jakub's profile and remove Rocket's Python directory from the PATH.

nannanli:
@John-A-Davies @plavjanik Hi John, Petr, 

Reading the discussion thread, it seems there is already a solution to the problem that Jakub encountered. As a follow-up, is the doc changes in this PR valid at this point? If not, do you think we could draft a troubleshooting topic to help people who might meet the same issue as Jakub? Thanks!

nannanli:
Hi all @plavjanik @John-A-Davies , I'm revisiting this PR. With 1.8.0 coming soon, are the changes in this PR still required or valid? Can we merge it or close it？ Thanks!

nannanli:
With 1.8 rolling out soon, closing this PR but feel free to reopen as needed. Thanks!

Issue #410, opened 2019-03-04
Update the installation package name for Zowe CLI


nannanli:
## Description

The name of the Zowe CLI installation package on doc site needs to be updated. The new name on zowe.org is `zowe-cli-package-1.0.0.zip` for 1.0.0 package and it will change with new release package uploaded. 

## Pages to Update
https://zowe.github.io/docs-site/latest/user-guide/cli-installcli.html#installing-zowe-cli-from-a-local-package

## Expected behavior
Like the Zowe package for z/OS install, might need to use a generic package name like `zowe-cli-package-<v.r.m>.zip` and explain the variable to reduce future maintenance effort.

JamesBauman:
@nannanli @jackjia-ibm - Hi Ashley/Jack -  

We dont mention the version# of the package in the above-referenced article (unless I missed it).   Currently, the DL media for CLI contains 1.0.1 in the file name.  Jack, is this on your radar to update the file name of the CLI installation file?  I presume "yes." but just want to be sure...

thx,
jb

jackjia-ibm:
@JamesBauman After released of v1.0.1, the download URL of CLI will be automatically updated from `zowe-cli-package-1.0.0.zip` to `zowe-cli-package-1.0.1.zip`. The doc mentioned `zowe-cli-bundle.zip`, which is the old download file name should be updated.

JamesBauman:
OK, I see it now @jackjia-ibm Jack. I'll go with @nannanli Ashley's suggestion to use generic references.

Thx

JamesBauman:
this will be published with the refresh PR #406

PR #412, opened 2019-03-04
updated the name of the CLI install package

JamesBauman:
Signed-off-by: bauja03 <james.bauman@broadcom.com>

- updated the name of the CLI install package.
- generic ref so that we dont have to update it every refresh

Close issue #410


JamesBauman:
merging updates now.....

Issue #401, opened 2019-02-28
Document ACF/2 and TSS steps for BPX.SERVER and BPX.DEAMON permissions for ZWESIS01 manual setup


Joe-Winchester:
## Description
1.0.1 introduced the need for the user running ZWESIS01 (by default IZUSVR) to have UPDATE ability for the BPX.SERVER and BPX.DAEMON FACILITY class, so that the program can run commands on behalf of different users (also known as impersonation).  This is described in the "manual setup for the APF server" in step 7,  "Security Environment Switching"

The steps are described for RACF, and don't include ACF/2 and TSS.  

## Pages to Update
install-zos.md  section 7 "Security Environment Switching"

nannanli:
Hi Mark @MarkAckert , I'm wondering if you know the steps to grant BPX.SERVER and BPX.DEAMON permissions for ACF/2 and TSS users that's mentioned in this issue. Could you advise? The section that talks about this for RACF users in the doc is https://zowe.github.io/docs-site/latest/user-guide/install-zos.html#manually-installing-the-zowe-cross-memory-server. 

Providing links to the ACF/2 and TSS documentation site is one option that is being considered. However, if any concrete commands or steps can be provided like what's done for RACF, that would be great. Thanks!

Joe-Winchester:
@ifakhrutdinov and @MarkAckert  - If we need anyone else to help here I met Julie Bergh from Rocket software at Share Phoenix and who is a WW Lead Security Advisor and has a lot of ACF/@ and TSS experience.  I also met Sai from Broadcomm at Share and she volunteered help if we need to reach outisde and "phone a friend" ?  Let me know and I can make the introduction/connection

ifakhrutdinov:
@Joe-Winchester thank you.

@MarkAckert, I think someone from Broadcom has already installed the cross-memory server, would that be possible to share the right ACF2/TSS commands?

MarkAckert:
I will work on a PR to add TSS/ACF2 commands. I suggest we also alter the documentation flow slightly. Right now the first thing users see are security commands to create the BPX facilities and grant UPDATE access, and they should probably first check if these facilities/permissions exist.

Sample Modified Flow:
<intro_bpx_description>:

To check if you have defined BPX.SERVER / BPX.DAEMON and IZUSVR (or STC user) has access, issue:
RACF:
TSS:
ACF2:

If the resources are not defined / user does not have access, follow these instructions to configure the facilities:
RACF:...
...

[...]

PR #437, opened 2019-03-22
Add TSS / ACF2 manual commands for BPX

MarkAckert:
Addresses #401

### DIFF
--- a/docs/user-guide/api-mediation/api-mediation-overview.md
+++ b/docs/user-guide/api-mediation/api-mediation-overview.md
@@ -24,7 +24,7 @@ Services that comprise the API ML service ecosystem are located behind a gateway
 
 ### Discovery Service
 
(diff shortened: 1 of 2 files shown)


nannanli:
Hi @MarkAckert @BrandonJenkins14 , this seems still a work in progress. Wondering if it's ready to go for v1.0.2?

JamesBauman:
Hi @nannanli Ashley - I chatted with Mark about this PR.  He would like to you review the PR.  Here are the notes from my chat with him:

i'd like ashley's review on the new format
because this block of text was incredibly confusing
the idea was
users may or may not have these resources/facilities already defined
and they should check first

but we leave everything up in the air
so i changed it to
first: check
second: if check failed, create
third: add new permissions
fourth: check to confirm
something like that
its good to go content wise from tss/acf2 security commands
my concerns are entirely on the digestibility of the doc

its complete otherwise
ready to merge
can go live whenever
sry not being on top this
jumping around different areas last week or two

Thx

nannanli:
@MarkAckert Hi Mark, 

 I'm looking at this PR that adds TSS/ACF2 manual commands for BPX: https://github.com/zowe/docs-site/pull/437. I made some edits and adjusted the format to reduce the level of list depth so it would be easier to read for users. See https://github.com/zowe/docs-site/blob/update-tss-acf2-bpx/docs/user-guide/install-zos.md#manually-installing-the-zowe-cross-memory-server.

For the task flow, users need to first check if they already have the facilities defined and UPDATE access granted. If not, they need to issue some commands to grant access. However, I noticed that RACF users have more steps to perform (Activate and RACLIST the FACILITY class, define BPX facilities) than ACF2 and TS users. I'm wondering if this is expected. Also could you take a look at the latest instructions and format and let me know if any issues?

PR #444, opened 2019-03-26
Troubleshoot Zowe Application Server

Joe-Winchester:
Signed-off-by: Joe-Winchester <winchest@uk.ibm.com>

### DIFF
--- a/docs/troubleshoot/troubleshoot-app-framework.md
+++ b/docs/troubleshoot/troubleshoot-app-framework.md
@@ -34,13 +34,13 @@ To solve the problem, follow these steps:
    When the desktop node server is able to communicate correctly, the line is followed by `- Ok`, so the log entry reads as follows:
 
    ```
-   ZIS status - Ok (name='ZWESIS_STD      ', zisRC=4, cmsRC=0,
+   ZIS status - Ok (name='ZWESIS_STD      ', cmsRC=0, description='Ok'
    ```
 
    If the line shows `Failure` such as:
 
    ```
-   ZIS status - Failure (name='ZWESIS_STD      ', zisRC=8,cmsRC-47,
+   ZIS status - Failure (name='ZWESIS_STD      ', cmsRC=39, description='Cross-memory call ABENDed'
    ```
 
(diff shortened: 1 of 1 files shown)


docs/troubleshoot/troubleshoot-app-framework.md
```diff
@@ -1,14 +1,55 @@
...
+   then, the setup and configuration of the cross memory server is completed successfully. You must follow the steps as described in [Manually installing the Zowe Cross Memory Server](../user-guide/install-zos.html#manually-installing-the-zowe-cross-memory-server) to set up the cross memory server.  
+
+   - Check that the ZWESIS01 started task is running and look into the log for any problems such as unable to find the load module.
+
+   If the line shows `Permission Denied` such as:
+
+   ```
+   ZIS status - Permission Denied (name='ZWESUS_STD    ',zisRC=8,cmsRC=33,
+   ```
+    - Check that the user ID of the ZOWESVR started task is authorized to access the load module. There is a security check in place to ensure that only authorized code is able to call ZWESIS01 as it is an APF-authorized load module. The setup for each security manager is different and documented in the section "Security requirements for the cross memory server".
```
vvvlc: I had another reason for 
```
ZIS status - Permission denied (name='ZWESIS_STD      ', zisRC=8, cmsRC=33, srvcRC=0, clientVersion=2)
``` 
In my case STC user assigned to `ZOWESVR` did not have `ZWES.IS`.

Joe-Winchester: @vvvlc - I've added a RACF specific note to cover this.  The instructions for the two other security managers Top Secret and ACF/2 are different and they are all covered in detail in the referenced chapter, however as you point out it's good to have the common problems covered in the mainline of the docs themselves.

It would be good to have a quick line in place for ACF/2 and TSS not to favour one security manager over another, but I'll have to defer to someone else more familiar with those security managers to add those two lines.  

***Note*** If you are using RACF then the problem may be that the user ID running the ZOWESVR started task, typically IZUSVR, does not have READ access to the ZWES.IS FACILITY class.

PR #465, opened 2019-04-05
Clean up Overview

BrandonJenkins14:
Draft PR where we can experiment with the overview page, which is a bit overwhelming in length and content. I began by reorganizing some text and headings, and introducing basic expandable sections for each component. More to do:

- I think that the introductory paragraph itself could use a re-visit to provide a compelling/concise explanation of the whole Zowe project
- Any other suggestions on how to make this section more "digestable"?
- Not sure if the Details html tag will do well or if we'll need another solution for expandables ( and javascript, or something?) @jackjia-ibm 
- Looking into adding a button look-and-feel to the expandable sections.

Signed-off-by: BrandonJenkins14 <brandon.jenkins@broadcom.com>


nannanli:
@BrandonJenkins14 This looks good to me. It is much improved and greatly reduces the length of the overview section. Do you want me to merge it?

PR #469, opened 2019-04-09
Move UI guidelines to the doc site

chentingcathy:
@jayenzi , 
I'd like to confirm with your about one change that I made: in [this topic](https://zowe.org/ui-guidelines-colors/), I changed **them-able** to **theme-able**, is that correct?
The Zowe color palette is designed and implemented in a **them-able** manner. 

Please feel free to review and revise. Thanks!


nannanli:
Thanks @BrandonJenkins14 for the edits and comments! All very helpful insights for this work. I held off the merge of this PR due to the conformance program for I thought this would be part of the program as a series of guidelines that extenders must follow so we might want to put more thoughts about the organization and location of the guidelines. However, I agree that we could merge this PR first and later iterate as more progress is made in the conformance program. 

@chentingcathy 
Hi Cathy, would you please take a look at Brandon's comment on the Embed Font section. The original input from Zowe.org is not very clear about the relationship of the three subtopics. By reading the content, I think Brandon's correct that "Standard", "Import", and "Specify in CSS" are three parallel sections within "Embed font". 

Another thing you might want to change is the size of the graphics. After build locally, the graphics looks big and they will be truncated when converted into PDF. So it will be great to adjust their size to be smaller. Otherwise, I agree, this PR is good to go.

nannanli:
@BrandonJenkins14 Perhaps we could bring our confusions to the Web UI conformance play forward meeting tomorrow and see if anyone could provide some insights.

nannanli:
Thanks Brandon for all the updates to move this forward! Only one last comment. I think these guidelines belong to the `Contributing` section on doc site. Like doc contributors, UI contributors and code contributors are primary personas for these guidelines when they contribute code changes or patches. Just moved it to the Contributing section but any other thoughts are welcome.

PR #471, opened 2019-04-11
Add Feedback link to doc site

nannanli:
Added the Feedback link to the nav bar to collect user feedback. 

Hi Jack @jackjia-ibm , wondering if you know a way to add this link to each page, right after the Propose content change in GitHub link?  This link goes to a documentation survey form. Thanks in advance!

### DIFF
--- a/docs/.vuepress/config.js
+++ b/docs/.vuepress/config.js
@@ -343,6 +343,7 @@ module.exports = {
       { tags: ['versions'] },
       {
         text: 'Feedback',
+        canHideFirst: true,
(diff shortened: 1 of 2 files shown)


nannanli:
Hi Jack @jackjia-ibm , that's a good point! We have many sections on navbar now and with more content coming in the future, we might have the possibility to add more sections. I'll open an issue and see how we can reorganize them. Thanks!

JamesBauman:
Looks good... nice addition to the docs site :+1:

PR #480, opened 2019-04-16
clarify prereqs not for CLI

BrandonJenkins14:
- Cleaned up some extra text in the CLI section to make the page easier to consume. 
- added "not required for CLI" to certain prereqs. Users are getting confused and thinking that they need Node on z/OS in order to use the CLI alone. 

Signed-off-by: BrandonJenkins14 <brandon.jenkins@broadcom.com>

### DIFF
--- a/docs/user-guide/systemrequirements.md
+++ b/docs/user-guide/systemrequirements.md
@@ -6,12 +6,10 @@ Before installing Zowe, ensure that your environment meets the prerequisites.
 - [Zowe Desktop requirements](#zowe-desktop-requirements)
 - [Zowe CLI requirements](#zowe-cli-requirements)
 
-## z/OS host requirements
-
-The following software is required for all Zowe components:
+## Common system requirements
 
 - z/OS Version 2.2 or later.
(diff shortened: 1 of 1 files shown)


nannanli:
Hi @BrandonJenkins14 , thanks for making the changes. It's much clearer now. Only one comment: 
I'm thinking if it would be helpful to further organize the requirements as follows by moving the other two z/OS requirements to the Zowe Desktop section and keep only the common requirements in the first section:

**Common system requirements**
- z/OS Version 2.2 or later.
- IBM z/OS Management Facility (z/OSMF) Version 2.2 or Version 2.3.

**Zowe Desktop requirements**
- **Node.js Version 6.14.4.1 or later *on the z/OS host* where you install the Zowe Application Server** 
- **IBM SDK for Java Technology Edition V8 or later**
- 833 MB of HFS file space
- Supported browsers

**Zowe CLI requirements**
(No change)

This way, Zowe CLI users don't have to watch for what's required and what's not and seems more straitforward IMHO. 

I'd also suggest that Joe @Joe-Winchester take a look at our changes here and also the overall sysreq section for I know the onboarding squad has been leading the install enhancements and they might have some good suggestion here. :-)

Issue #502, opened 2019-04-29
Update doc to cater for nodeCluster instead of nodeServer


John-A-Davies:
## Description

The doc currently describes the situation where `nodeServer` is used.  But from `nodeCluster` is used from now on.
## Pages to Update

Page 145 and others.

## Expected behavior

The doc describes the situation where `nodeCluster` is used.  
## Additional context

A defect was found ( https://github.com/zowe/zlux-server-framework/issues/88) by using AppScan.  The solution to this problem was to use `nodeCluster` instead of `nodeServer`.  

Any changes in Zowe runtime behaviour would best be confirmed by @1000TurquoisePogs

nannanli:
PR #503 fixed this. Closing.

PR #503, opened 2019-04-29
Node server to node cluster

John-A-Davies:
Replace wording about `nodeServer` with `nodeCluster`.


John-A-Davies:
Fixes issue #502.

nannanli:
Hi John, @John-A-Davies 

The following occurrences were also found. Wondering if they should also be replaced. Could you take a look? 
cc @1000TurquoisePogs 

1. docs-site\docs\extend\extend-desktop\mvd-dataservices.md (1 hit)
- Line 71: An example is available in **`sample-app/nodeServer/ts/helloWorld.ts`**

2. docs-site\docs\extend\extend-desktop\mvd-plugindefandstruct.md (4 hits)
  - Line 24: #### **nodeServer**
  - Line 26: When an application plug-in has router-type dataservices, they are interpreted by the Zowe Application Server by attaching them as ExpressJS routers. It is recommended that you write application plug-ins using Typescript, because it facilitates well-structured code. Use of Typescript results in build steps because the pre-transpilation Typescript content is not to be consumed by NodeJS. Therefore, keep server-side source code in the **`nodeServer`** directory. At runtime, the server loads router dataservices from the `lib` directory.
  - Line 38: The `lib` directory is where router-type dataservices are loaded by use in the Zowe Application Server. If the JS files that are loaded from the `lib` directory require NodeJS modules, which are not provided by the server base (the modules `zlux-server-framework` requires are added to `NODE_PATH` at runtime), then you must include these modules in `lib/node_modules` for local directory lookup or ensure that they are found on the `NODE_PATH` environment variable. **`nodeServer/node_modules`** is not automatically accessed at runtime because it is a dev and build directory.
- Line 54: **location**: This is a directory path that is relative to the server's executable (such as **`zlux-app-server/bin/nodeServer.sh`**) at which a `pluginDefinition.json` file is expected to be found.

3. C:\GitFolder\docs-site\docs\user-guide\mvd-configuration.md (1 hit)
 - Line 407: - Zowe Application Server: **`zlux-app-server/log/nodeServer-yyyy-mm-dd-hh-mm.log`**

John-A-Davies:
Yes, all these instances of `nodeServer` should be retained and not changed to `nodeCluster`.

PR #527, opened 2019-05-13
Button for CLI Ref download

BrandonJenkins14:
CLI reference guide is in-demand, but is easily missed. Add a button so that the download is more prominent.

Consider future changes to home page design - how can we make more content visible without scrolling, or highlight popular topics, without cluttering the page?


JamesBauman:
merging to master to publish

Issue #498, opened 2019-04-26
Adjust pipeline to only validate branch level broken links.


jackjia-ibm:
## Is your request for enhancement related to a problem? Please describe.

Right now every build will validate the whole documentation including all archives.

## Describe the solution you'd like

Since we have several archived version now, and the validation will take longer time, it's better to just validate broken links of the branch is working on.

jackjia-ibm:
PR https://github.com/zowe/docs-site/pull/531 created.

PR #531, opened 2019-05-14
issue #498 perform link check only on current build

jackjia-ibm:
Since we have multiple archived versions now, the test stage now takes 12 min.

This change reduce the test stage to 3 min by skipping checks on archived versions. A new build parameter `FULL_SITE_LINKS_CHECK` is added in case we want to test the whole website.

Note: with this changes, the link check test will not check external links.


nannanli:
Looks good to me. Leaving this open for a while to see if there are any other feedback.  Otherwise, it can be merged. :)

PR #542, opened 2019-05-15
Replace placeholder credentials for maven guide with real ones

IgorCATech:
Update Maven guide


IgorCATech:
I've just noticed this docs/user-guide/install-zos.md file in the PR, apparently there's a change, but one I'm unaware of. @BrandonJenkins14 @nannanli @JamesBauman  Reviewers, please let me know if you recognize this change. If not, before merging, take it off.

Issue #550, opened 2019-05-16
Move code guidelines from Zowe.org to the doc site


nannanli:
Move the code guidelines here https://zowe.org/code-guidelines/ to the Zowe doc site like what's done with the UI guidelines https://zowe.github.io/docs-site/latest/contribute/guidelines-ui/ui.html#clear.

BrandonJenkins14:
I thought I'd work on this one since the UI guidelines just came over too. 

I am curious though - what is the distinction between Contribute and Extend sections? To me they are really the same thing. As I look at these code guidelines/ui guidelines, they are quite relevant to developing new components, which is described in "Extending". The only outlier is really "Contributing to documentation", but I would argue that it would be OK to have that as part of Extending as well. 

I think a single nav section would work for this - then we'd just have to decide on one name or the other "Contributing" or "Extending". I think that "Contributing" is the term that most people are familiar with in relation to open source software. 

What do you think @nannanli ?

PR #553, opened 2019-05-20
Move code guidelines to docs-site

BrandonJenkins14:
I moved the "Code Guidelines" from Zowe.org to docs-site for #550 . I copied the content, added to TOC under Contributing, and spent time reviewing/adding to the text. **Please review the content itself and make changes as you see fit**. 

Not quite ready to merge, first some questions to answer -

- There seems to be some inconsistency in regards to which headings are bolded in the left-hand TOC when first opening each navigation section. Sometimes the top topic is highlighted by default, other times there are multiple topics highlighted by default. I haven't found a pattern yet. Please build and let me know if you notice the same issue. 

- I am not sure that Contributing is the place for this, or that we need both Contributing and Extending sections in general. I will do some experimentation in another branch and see if combining these sections with our tutorials under one "Contributing" section would make sense. Will open another PR for discussion.

### DIFF
--- a/dco_signoffs/Adam-Battenburg-docs-site.txt
+++ b/dco_signoffs/Adam-Battenburg-docs-site.txt
@@ -1,0 +1,52 @@
+I, Adam-Battenburg hereby sign-off-by all of my past commits to this repo subject to the Developer Certificate of Origin (DCO), Version 1.1. In the past I have used emails: Adam.Battenburg@ibm.com
+
+27714fd7df14d72a463692e750d3c73f5c9ab673 Fixed build errors by fixing toc
+a5c7ead9c908b7d04eba5e36445f7c1ed846fd2d Add google site verification to head.
+d4172dda8984f1ddee978221bb06b209471a0fa5 Fix broken link
+164fd1bcc5737af7fb9e3d0c1c60d54a744b6c87 Moved work in progress to another branch
+81c5abc488f3c392aa90df1193b4690aff58a3f2 Changes to webui tutorial.
+946f2e2a3a0cc54ff2929d01d4021a11e991d9e9 Created new folder for work in progress content
+cb861bdd3181e06d5daf4988c383fa1fcaeda7fb Merged header-concept and changed samples repo links.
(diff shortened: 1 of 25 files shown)


nannanli:
Hi Brandon, thanks for making the changes! For the two questions you have, here are my thoughts/findings: 
1. Not sure if it's the same screen you saw. Looks as expected on my side. :)
WeChat Screenshot_20190521143134

2. For this one, Contributing and Extending have some slight differences as I understand. 
    - Contributing is for various contributors. They could simply fix a bug and open a pull request or add new features to Zowe. Therefore, they need some guidelines or recommendations on how to work with code, UI, doc, etc to get involved. It's not "how to develop and contribute plug-in to Zowe" but instead "what I could do and how to do it to participate in this project". 
    - Extending is for the plug-in developers, kind of a developer guide that talks about how to develop on Zowe. Extenders are not necessarily contributors. They could develop a plug-in for use internally within their enterprise or choose to ontribute it back to the community. 

   Therefore, they are two sepatae sections at first. However, we coud always look at solutions to make this much clearer.

JamesBauman:
Hi @nannanli Ashley - 

Re point 1, i believe all the top level headings in the TOC should be bold - in this case, the heading UI Guidelines is not.  on a related note, I am observing similar behavior in other sentions; not just contributing, See https://zowe.github.io/docs-site/latest/getting-started/overview.html and https://zowe.github.io/docs-site/latest/user-guide/installandconfig.html.   I'm not sure what is the solution, perhaps its the syntax in the config.js file (??).

jackjia-ibm:
@BrandonJenkins14 I made a change to disable the bold on sidebar. Please see if this is as you expected.

jackjia-ibm:
@BrandonJenkins14 I see. I've reverted the change.

Just to clarify, in this picture
Screen Shot 2019-05-21 at 1 59 23 PM,
there are two "Release Notes", do you mean the first "Release Notes" text? I think vuepress render it as a sidebar group, it's not clickable like the second one.

If that's correct, we need to change the way how sidebar group is rendered. I agree it looks more complicated, a separated issue will be better to track it.

PR #564, opened 2019-05-23
Make command examples in quick start guide more consistent

tjohnsonBCM:
Attempted to make the order and sample values of params for the command examples more consistent. Not sure however which values should be preferred for consistency, like username for example (`user123`, `myusername123`, or `ibmuser`).


JamesBauman:
@BrandonJenkins14  = This is already merged... nothing for me to approve...

PR #579, opened 2019-05-30
Remove /var/zowe references

Joe-Winchester:
Remove three references to /var/zowe from the install docs

### DIFF
--- a/docs/user-guide/install-zos.md
+++ b/docs/user-guide/install-zos.md
@@ -176,7 +176,7 @@ To install Zowe API Mediation Layer, Zowe Application Framework, and z/OS Servic
 
     ```yaml
     install:
-     rootDir=/var/zowe/1.3.0
+     rootDir=/tmp/zowe/1.3.0
 
     api-mediation:
       catalogPort=7552
--- a/docs/user-guide/install-zos.md
+++ b/docs/user-guide/install-zos.md
@@ -160,7 +160,7 @@ To install Zowe API Mediation Layer, Zowe Application Framework, and z/OS Servic
 
 2. Review the `zowe-install.yaml` file which contains the following properties:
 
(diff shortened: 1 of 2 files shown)


docs/user-guide/install-zos.md
```diff
@@ -176,7 +176,7 @@ To install Zowe API Mediation Layer, Zowe Application Framework, and z/OS Servic
 
     ```yaml
     install:
-     rootDir=/var/zowe/1.3.0
+     rootDir=/tmp/zowe/1.3.0
```
rpenny125: why wouldn't you suggest /usr/lpp/zowe/1.3.0?

Joe-Winchester: For SMP/E we are going to use /usr/lpp/zowe/v1 and update this between versions, so we don't keep 1.2 and 1.3 and 1.4 separate, rather we always replace /usr/lpp/zpwe/v1.  I didn't want to suggest this before we have SMP/E working as we don't want customers using the simple zowe-install.sh to push this to /usr/lpp/zowe/v1 as 1) they probably won't have permission and 2) we want that reserved for SMP/E so that the CSI installs into that location and doesn't want anyone else to use that.

nannanli:
@Joe-Winchester Any comments on the feedback above? Thanks!

Joe-Winchester:
@nannanli - the reason I don't want us to tell customers to use /var/lpp is because is what we are going to reserve for the SMP/E install that we are still building.   When that is available the docs will be updated, hopefully in July.  The current docs describe how to install from the unix shell script when we don't want to encourage people to target /var/ which is why for the PR I change ```/usr/``` (which as Rober said is wrong) and modified to ```/tmp/``` which feels safer.

PR #587, opened 2019-06-06
Update cli-extending-a-plugin.md

charlotteliu:
@brightside/imperative are no longer in package.json,   use @zowe/imperative


ChrisBoehmCA:
That's right Brandon

PR #601, opened 2019-06-21
Diagram update

Joe-Winchester:
Update of the architecture diagram.  New .png as well as the original .pptx


nannanli:
Thanks Joe for the update! @Joe-Winchester As Brandon mentioned, merging this to the master branch can publish the change asap. Just click the **Edit** button besides the PR title, and then select `master` from the base branch list, it will work. :)

solsu01:
The diagram looks great. Should the CLI and plugins also be colored green to indicate that they're node?

Issue #618, opened 2019-07-11
Nav bar breaks at certain widths


BrandonJenkins14:
Top nav bar on Zowe Docs Site breaks. The drop-downs menus fall below search bar.

Tested in Chrome; seems to break between 960px and 959px width.

![image](https://user-images.githubusercontent.com/36675406/61061249-e1967a00-a3c9-11e9-82e8-b1525a626b36.png)

Issue #285, opened 2018-12-07
Search box overlaps Zowe 


JamesBauman:
## Description
When you reduce the window size (not Tablet mode) the search box in the menu bar overlaps Zowe. 
## Pages to Update
NA

## Screenshots
See attached.

## Expected behavior
When i reduce the window, the menu bar should switch to Tablet view just as the search box is about to overlap Zowe. 

## Additional context
NA
![capture](https://user-images.githubusercontent.com/36677566/49663374-6f694b80-fa1c-11e8-8f2c-c713098e8205.PNG)

jackjia-ibm:
@JamesBauman Is this still an issue?

PR #624, opened 2019-07-15
update convert point to wrap navbar

jackjia-ibm:
Signed-off-by: Jack (T.) Jia <jack-tiefeng.jia@ibm.com>

To fix issue #618.


jackjia-ibm:
This should also fix the search box overlap addressed in https://github.com/zowe/docs-site/issues/285. Please confirm, thanks.

PR #625, opened 2019-07-17
update doc to match install template change

stevenhorsman:
Signed-off-by: stevenhorsman <steven@uk.ibm.com>

### DIFF
--- a/docs/user-guide/install-zos.md
+++ b/docs/user-guide/install-zos.md
@@ -165,7 +165,7 @@ To install Zowe API Mediation Layer, Zowe Application Framework, and z/OS Servic
 
 2. Review the `zowe-install.yaml` file which contains the following properties:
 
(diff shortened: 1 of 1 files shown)


nannanli:
@stevenhorsman Hi Steve, is this a what's new item for the upcoming release? I'm trying to understand if we need to call out the install template changes in the release notes. Thanks!

stevenhorsman:
Hey, this is all part of the CUPIDS re-work. Don't call it out this release in the release notes as we've made a lot of foundational changes to set up the future work, but tried to avoid disrupting current behaviour. The plan is then that in the next 1/2 releases we will be dropping the big change content, like SMPE installer and supporting componentisation and selective start-up of zowe parts, so those will need a lot more doc and shouting about in release notes. I hope that makes sense!

nannanli:
It makes sense. Thanks Steve!

PR #626, opened 2019-07-17
Expand trouble shooting section to include detail L2 support should understand

Colin-Stone:


Colin-Stone:
> Hi @Colin-Stone, I made some changes and moved some things around. Let me know if you are OK with the changes.

Thanks Brandon yes it looked fine to me. I reckon we should hit the DCO button and merge it

PR #657, opened 2019-08-15
Added back description about node.js v8 support

nannanli:


MarkAckert:
Can we recommend that users install Node v8, rather than Node v6? While we still support Node v6 right now, v6 is EOS except for paying customers. @Joe-Winchester

Issue #654, opened 2019-08-13
Add navigation from zowe.org to OMP conformance program page


supmi01:
## Describe the solution you'd like
Proposal to add navigation from zowe.org to OMP Conformance program page https://www.openmainframeproject.org/projects/zowe/conformance so that zowe users can more easily access the information about what extensions are conformant. 
The most logical placement to me is the footer of the https://zowe.org/home/ where there is a link to OMP. All this should be done in the spirit of not propagating any vendor or company.

ecgdave1:
Would be nice to see it on the zowe.org nav bar.  Visibility to conformant tools illustrates value.

IgorCATech:
I agree with ecgdave1. In the footer it would remain unnoticed.

nannanli:
Agree to add a `Zowe Conformance` link in the zowe.org nav bar.  That page is under review so we could revisit this issue to update the community website when the page goes alive. 
cc @Tbr00ksy

Tbr00ksy:
I agree with all of the above. The conformance site should go live on August 20th, at which point we can add a button/link to the zowe.org home page, and include it in the extending section of the docs site as well.

supmi01:
Please also check the right naming. I am perhaps wrongly using the name OMP Conformance Program Page, but what it should be is Zowe Application Directory / Marketplace. @Tbr00ksy will know this for sure.

Tbr00ksy:
Hi @supmi01, the "Zowe Conformance Program" will be the name of the page on the OMP website. On that same page there will be a section called the "App Directory" that will list the conformant apps. We are modeling it after CNCF: https://www.cncf.io/certification/software-conformance/

For any feedback on the design/content please see this issue: https://github.com/zowe/community/issues/234

nannanli:
Quick update about this issue: 
- The link to the Conformance program is alive on [zowe.org](zowe.org) now (both nav and footer section).
- Created a PR to add the link to the extending section on the doc site. See #669 . Will close this issue after the merge.

nannanli:
The Zowe Conformance Program link is also available on the doc site now. See the nav under Extending. https://zowe.org/docs-site/latest/extend/zowe-conformance-program.html

PR #669, opened 2019-08-20
Add Zowe Conformance Program link #654

nannanli:


nannanli:
@BrandonJenkins14 Agree. That would be helpful to keep users updated of the doc site changes.

Issue #670, opened 2019-08-20
Incorrect branding: CA ACF2 and CA Top Secret


JamesBauman:
See the topic:  https://zowe.org/docs-site/latest/user-guide/systemrequirements-zosmf.html#configuring-z-osmf-2

ACF2 and Top Secret names branded incorrectly. 

Change all references of CA Technologies ACF2 and CA Technologies Top Secret to CA ACF2 and CA Top Secret respectively.

JamesBauman:
Content updated....

PR #671, opened 2019-08-20
Acf2 tss iss670

JamesBauman:
Removed "Technologies"  from the names of the CA security products reverenced in the article.
Updated the target URLs to point to the techdocs versions of the URLs. 

Resolves issue #670


JamesBauman:
Ready to publish ...

Issue #522, opened 2019-05-09
Adopt VuePress v1.0 once it's official released


jackjia-ibm:
## Is your request for enhancement related to a problem? Please describe.

Current VuePress v0.14.1 which is a little out-dated. Migrating to v1.x will give us more flexibilities to apply plugins to the website. The new plugin feature is introduced by v1.x of VuePress.

This task may require us to revisit some customizations we made related to css styles, versions dropdown, icon links, copy-to-clipboard, etc.

## Additional context

Current result of `npm view vuepress` is:

```
latest: 0.14.11       next: 1.0.0-alpha.47  
```

We need to wait for the official release to avoid duplicated re-eject work.

jackjia-ibm:
PR #672 resolved the issue.

PR #672, opened 2019-08-20
Issue #522 migrate to Vuepress v1

jackjia-ibm:


jackjia-ibm:
This PR should solve issue #522.

Vuepress had changed the `Last Updated` to the bottom of the page, I found it's not quite easy to bring it to top as before, so left it there.

This should also solve all npm audit warnings.

PR #693, opened 2019-08-30
Fix a format issue

nannanli:


nannanli:
@BrandonJenkins14 The build passes now. Guess there was an issue with the build when we were updating the site URL. ：）

PR #694, opened 2019-08-30
Zowe smpe reorg

John-A-Davies:
"Given this is the first release of the SMP/E package, thus, no PTFs have been created. " is a slight tautology; I suggest  "Since this is the first release of the SMP/E package, no PTFs have been created. "

the text ___usr/lpp/zowe___ is the name of the mountpoint should start with a slash, i.e. ___/usr/lpp/zowe___ is the name of the mountpoint 

The text "or zFS paths or zFS paths " is repeated.

In this text "mount the new file system at IP time. This action can be helpful if an IP occurs  ",  IP should be IPL.

### DIFF
--- a/docs/user-guide/install-zowe-smpe.md
+++ b/docs/user-guide/install-zowe-smpe.md
@@ -124,7 +124,7 @@ All issues of previous releases of Zowe that were resolved before August 2019 ha
 
 ### Service level information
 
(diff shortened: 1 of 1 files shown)


Joe-Winchester:
smpe-file merged manually into zowe-smpe branch

PR #697, opened 2019-09-05
Publish the Active Development Documentation

BrandonJenkins14:
- New branch accessible via version-picker
- New Release Notes topic specific to this version
- New installation, quick-start, plug-in procedures
- MQ and IMS plug-ins added 
- TPSRs and BoM

Do not merge until package available on Zowe.org


jackjia-ibm:
@BrandonJenkins14 I see all broken links are in the TPSR page.

If we merge to `active-development` branch, these failures will be carried over and you won't see a successful active-development build. Since the pipeline stops at the broken link check, any new changes to the branches won't be published. So you will not see website updated.

If we decided to merge, we need to apply some tweaks to ignore the error and continue on publishing. @MarkAckert what do you think?

jackjia-ibm:
If we could prepend `http://` to the links, will solve the issues. The failed links are not recognized as external URLs but as sub folders under our docs domain. The link checker doesn’t really go through third party links except for our own domains.

PR #710, opened 2019-09-18
Fix mini-ToC in the summary of changes article

JamesBauman:
summary of changes contained a link to the title of the article... removed


JamesBauman:
applied ifx

PR #720, opened 2019-09-26
add background about branches and add step to switch branches

BrandonJenkins14:
Describe why there are different versions of CLI, corresponding sample branches, and the benefit of developing for both. 

Also add step to switch to `lts-incremental` branch for this version of tutorial. 

Signed-off-by: BrandonJenkins14 <36675406+BrandonJenkins14@users.noreply.github.com>

### DIFF
--- a/docs/extend/extend-cli/cli-setting-up.md
+++ b/docs/extend/extend-cli/cli-setting-up.md
@@ -27,7 +27,7 @@ Follow these steps:
 1. `cd` to your `zowe-tutorial` folder.
 2. `git clone https://github.com/zowe/zowe-cli-sample-plugin`
 3. `cd` to your `zowe-cli-sample-plugin` folder.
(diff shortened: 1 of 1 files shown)


docs/extend/extend-cli/cli-setting-up.md
```diff
@@ -19,8 +27,9 @@ Follow these steps:
 1. `cd` to your `zowe-tutorial` folder.
 2. `git clone https://github.com/zowe/zowe-cli-sample-plugin`
 3. `cd` to your `zowe-cli-sample-plugin` folder.
-4. `npm install`
-5. `npm run build`
+4. Switch your local branch to `lts-incremental`
```
zFernand0: Since we already provide a `command-by-command` instruction, I suggest changing this statement to `git checkout lts-incremental`

PR #764, opened 2019-10-15
Reduce repetitive information

nannanli:

The verify build information is the same as that on zowe.org. Also we point users to the post-download page to fetch some files, so it makes sense to leave the information in only one location. The repetitive info on doc site is being removed now. Instead, a link is provided for users to get the instructions.


nannanli:
@jackjia-ibm Hi Jack, see my comments in the description section. Would you please take a look and let me know if any comments? Thanks!

PR #774, opened 2019-10-17
node with 8.16.1 warning and 8.16.0 link

Joe-Winchester:
Signed-off-by: Joe-Winchester <winchest@uk.ibm.com>


nkocsis:
There is a link to the old version of Node (8.16.0) that we can direct customers to if needed. You can find the details in issue 142

https://github.com/ibmruntimes/node/issues/142

Joe-Winchester:
Tks @nkocsis for finding the 8.6.0 mistake that should have been 8.16.0.  I've pushed an update.  

@IgorTodorovskiIBM - for `export __UNTAGGED_READ_MODE=V6` I assume that once there is an update to 8.16.1 on the download site, we will be able to remove this update from the doc, and instead we'll be able to add the export statement to our shell script that initializes our USS environment and all will be well ?  If so then yes as soon as there is an 8.18.1 update we'll update Zowe and we'll also have to describe how to manually apply the export for any older versions of Zowe folks are running 8.16.1 with.   Do you have a time frame for when a fix might be available for 8.16.2 ?

PR #780, opened 2019-10-21
Propagate the Node v8.16.1 warning change to 1.5.x

nannanli:

Fixes: #774 

In v1.4.x doc, we stated support of node V8.


Joe-Winchester:
Good job picking this up @nannanli - I didn't think to consider that we only recently switched to node 8.  Thank you

PR #797, opened 2019-10-24
Update active dev from master

BrandonJenkins14:
I'll do this from time-to-time so that active development doesn't fall behind.


jackjia-ibm:
@BrandonJenkins14 for DCO, I think we have to do similar like what we did yesterday, mark DCO check passed and update dco_signoffs for this release in other PRs.

JamesBauman:
@MarkAckert - how should we proceed with this PR?  Given the recent conversations about DCO. I fear that squash merge is the appropriate action. It appears that the unsigned commits originated from the master branch that we pulled into this branch. 

Copying @BrandonJenkins14.

thx,
jb

PR #815, opened 2019-11-11
Separate editor doc to include it in Editor app

jasonenglish2039:
Issue number: N/A


nannanli:
Looks good to me. Approved. 

Only one comment is that shall we also separate the "Using the Workflows application plug-in" topic for it is similar to the using Editor one. They both talk about details of using a desktop application plug-in.

jasonenglish2039:
I know it is a bit awkward that we're separating the one app but not the others. I will separate workflows in another PR.

PR #844, opened 2019-11-20
Update active dev

BrandonJenkins14:


MarkAckert:
Set DCO to pass, these missing DCO's are historical at this point - already merged in master, and should be attributed in the dco_signoffs folder.

PR #852, opened 2019-11-27
Rename members

stevenhorsman:
Updating samplib member names.

Issue number: https://github.com/zowe/zowe-install-packaging/issues/864

### DIFF
--- a/docs/troubleshoot/app-framework/app-known-issues.md
+++ b/docs/troubleshoot/app-framework/app-known-issues.md
@@ -69,7 +69,7 @@ To solve the problem, follow these steps:
     ```
     Check that the user ID of the ZWESVSTC started task is authorized to access the load module. Only authorized code can call ZWESIS01 because it is an APF-authorized load module. The setup for each security manager is different and is documented in the section "Security requirements for the cross memory server" in the topic  [Manually installing the Zowe Cross Memory Server](../../user-guide/install-zos.html#manually-installing-the-zowe-cross-memory-server).
 
-    **Note** If you are using RACF security manager, a common reason for seeing `Permission Denied` is that the user running the started task ZWESVSTC (typically IZUSVR) does not have READ access to the FACILITY class ZWES.IS.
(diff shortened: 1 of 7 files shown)


stevenhorsman:
I've updates the doc for the cancel commands. I'm pretty sure the RDEFINE is based on the PROC name, not the job name - that's what Onno's ZWESECUR and the zowe-config-stc script uses

John-A-Davies:
Yes @stevenhorsman you are correct.
The doc says 
```
The START command specifies the member name to start and the job name to use. The member name is the name of a member of a partitioned data set that contains the source JCL for the task or job to be started. Using the STARTED class, RACF can assign different user IDs and group names to the same started member, depending on the job name that is used. CICS® can use this, for example, to allow one procedure to be used for various differ-ent CICS regions, which might have different security requirements.
```
The jobname *can* be used to select userid and group, but you have to specify it on the STARTED class definition.
```
Resource names in the STARTED class are of the form membername.jobname; for example, CICS.JOBA, CICS.REGION2, or IMS.PROD. The resource name is of the form membername.membername if no jobname is provided.
```
and, since we have `STARTED    ZOWE*.* (G)` , the jobname is * so it just matches the membername.

Conclusion:  The userid and group chosen to run the STC that you start is determined by what's in the matching RACF STARTED profile definition; `membername.jobname`.

stevenhorsman:
Done - good suggestion

stevenhorsman:
Made requested updates and modified ZWES entries to ZWEX to reflect the changes in zowe-install-apf-server

nannanli:
Hi Steve @stevenhorsman , are the rename changes targeted for the v1.8.0 release? Just want to confirm. Thanks!

stevenhorsman:
@nannanli - Hi - yes the rename members (along with lots of other changes) are planned for 1.8. We've got a few doc tasks around, but will have to make a concerted effort to review them all before release to make sure we haven't missed anything.

PR #883, opened 2019-12-31
Reintroduce doc for CLI prompt feature

BrandonJenkins14:
zowe/zowe-cli#269 added a new feature to CLI so that users can mask values when typing options on their screen. 

The doc was published, but has since mysteriously gone missing. This PR is a recreation of that content from old drafts/notes. 

Would appreciate a quick SME review to ensure accuracy before merging into 1.8 release doc.


tjohnsonBCM:
I think the absence of documentation for this feature may have been deliberate. The `prompt*` feature only exists in the @latest version of Zowe CLI, not in @lts-incremental. So I think this should only be documented in the Active Development branch.

PR #899, opened 2020-01-09
Fix TPSRs

BrandonJenkins14:
zowe/zowe-dependency-scan-pipeline#2 

Manually updated 4 licenses that are listed incorrectly for 1.7.0 release. 

Automation must be corrected to avoid future maintenance and inaccuracies.


nannanli:
@BrandonJenkins14 Hi Brandon, sorry for not commenting earlier. There are still 26 occurrences of `UNKNOWN` in TPSR. I seems those need to be corrected as well. For example, 
Screen Shot 2020-01-13 at 3 17 14 PM
How can we add them to the automation pipeline?

PR #920, opened 2020-01-14
Add scs lts

BrandonJenkins14:
Secure Credential Store Plug-in for Zowe CLI, to be published as part of Zowe v1.8.0.  

Two other PRs are open to publish into Active Development and the future "v1.10.0" doc. 

DO NOT MERGE until plug-in is available on Zowe.org package.


zFernand0:
Since The SCS plugin was donated only with the `@latest` compatible version in mind, I believe we can close this PR (without merging).

PR #1000, opened 2020-02-18
Node.js pre-req formulation change

IgorCATech:
Current Node.js pre-req formulation doesn't reflect the requirement precisely enough. We need to specify that we support LTS versions starting from Node v12. Also, a list of supported Node version is added.

#### Developer's Certificate of Origin (DCO)

Signed-off-by: Igor Kazmyr <43951184+IgorCATech@users.noreply.github.com>


IgorCATech:
Good call, @BrandonJenkins14 ! Thank you :)

PR #1006, opened 2020-02-19
Update logo

BrandonJenkins14:
Adding new Zowe Logo to Docs Site for v1.9.0 

Something could probably be done to make it more visible in the nav or more prominent on homepage, but not sure I have the time to work with that right now. 

New Logo in nav
image

Old Logo in nav
image


nannanli:
> Should this be ported to all current/archived branches?
@BrandonJenkins14 I think so. It would be confusing when users see different logos after they switch versions.

PR #1080, opened 2020-03-16
Correction on MFA support

1000TurquoisePogs:
MFA works with or without ZSS. The MFA section of the documentation explains the choices, but only after making the statement that it always uses ZSS, which is not true.

Signed-off-by: 1000TurquoisePogs sgrady@rocketsoftware.com


1000TurquoisePogs:
The diff is hard to read. All I did was removed
" or by accessing the zss `/login` REST API endpoint. 
All of these methods of authentication are backed by calls to SAF VERIFY through ZSS. This means that the App-Framework can work with any MFA product that supports pass-through of the MFA credentials as the passphrase used in SAF VERIFY."

Issue #1073, opened 2020-03-05
[User feedback] Clarify how to configure ICSF for Zowe


nannanli:
## Is your request for enhancement related to a problem? Please describe.

Feedback: Regarding Zowe installation guide, ICSF related section is too simple and too general, not so helpful for end user to install Zowe. 
 
## Describe the solution you'd like

Could you please add more detailed steps and description on how to configure that for Zowe installation to make it more user-friendly?

## Related doc pages

https://docs.zowe.org/stable/user-guide/install-nodejs-zos.html#hardware-and-software-requirements

## Additional context

ICSF is a prereq for Node.

nannanli:
One option is to add links to the relevant ICSF books.

NayerNajafi:
Hi @nannanli 
Here is the basic information about ICSF and why it should be available for Zowe in Z/OS:

* ICSF is a base element of z/OS and is included in z/OS package. It is not a Priced Feature. Each release of z/OS includes a particular ICSF FMID level as part of its base. For example, ICSF with FMID HCR77C0 is the base element of z/OS V2R3. 

* By default its SMP/E installed data sets has HLQ: CSF

* ICSF is not a prerequisite for Zowe, It is a prerequisite for IBM SDK for Node.js - z/OS. Since Zowe Application Framework (Zowe Desktop) requires Node.js, ICSF is a must.

* ICSF for Node.js requires neither Cryptographic Hardware nor Crypto cards.

NayerNajafi:
@nannanli 
Here is some sort of minimal configuration to set up ICSF:

1- Add CSF.SCSFMOD0 to the LNKLST concatenation.

2- APF Authorize CSF.SCSFMOD0. 

3- Modify IKJTSOxx to add CSFDAUTH and CSFDPKDS in the AUTHPGM parameter list and in the AUTHTSF parameter list.

4- Create the following ICSF Key Data Sets:
- Cryptographic Key Data Set (CKDS) (Sample JCL: SYS1.SAMPLIB(CSFCKDS))
- Public Key Data Set (PKDS) (Sample JCL: SYS1.SAMPLIB(CSFPKDS))
- (Optional) If you desire to have PKCS#11 support through ICSF, you should create Token Key Data Set (TKDS) (Sample JCL: SYS1.SAMPLIB(CSFTKDS))

[...]

PR #1100, opened 2020-04-02
Improve the Node.js install doc

nannanli:
- Added the high-level steps for configuring ICSF. It's a prereq for node.js.
- Removed out-of-date download link.
- Refactored the whole topic to better organize the information.

Issue number: #1073


nannanli:
@NayerNajafi @Joe-Winchester Hi Nayer, Joe, 

This PR addresses issue #1073 about adding more details about configuring ICSF which is a prereq for node.js. 

I added the high-level steps in our doc and also points users to the ICSF doc for more details. Although ICSF itself is a prereq for Node, but what we do now is to pull most node related information in the Zowe doc to reduce the redirection between Zowe doc and node doc and also confusion between what node prereqs versus what Zowe actually requires (like Python, make is not required). Hope this would help clarify a lot of confusion around node install. 

Would you please review this PR and let me know if any comments or suggestions? Thanks!

Joe-Winchester:
Looks great - thank you @nannanli and @NayerNajafi

PR #1114, opened 2020-04-09
updates to 1.10.0 release notes

ajinrocket:
Signed-off-by: alexanderjin <ajin@rocketsoftware.com>

Issue number:


docs/getting-started/summaryofchanges.md
```diff
@@ -34,30 +34,41 @@ The following new feature was added to the Zowe API Mediation Layer in this vers
 
 - Zowe API ML can now use z/OSMF to provide JSON Web Tokens (JWT). [#433](https://github.com/zowe/api-layer/issues/433)
 
+#### ZSS
+
+- Fast EBCDIC to UTF8 character translation is now supported by using the TROO instruction with a "EBCDIC 1047 to ISO/IEC 8859-1" translation table. [#127](https://github.com/zowe/zowe-common-c/pull/127)
+- Performance improvements in character conversion, JSON and collections code. [#162](https://github.com/zowe/zss/pull/162)
+- The code now prints fewer warnings when attls is not set up. [#130](https://github.com/zowe/zowe-common-c/pull/130)
+- ZSS logs belonging in the ZSS repo have been refactored so that they now use the Zowe logger and message IDs. [#163](https://github.com/zowe/zss/pull/163)
+- Support for APIML JWT tokens sent via browser cookies or Authentication: Bearer has been implemented. Once the token is validated, a cookie with the same name is sent back and creates a ZSS sesstion token for compatibility with ZSS auth. [#132](https://github.com/zowe/zowe-common-c/pull/132)
```
1000TurquoisePogs, line 43: Technically true but only in the case that the key is configured properly and we do not yet do that, but will in 1.11
I'd omit this for that reason, don't want to make people think we're finished.

Issue #991, opened 2020-02-17
question: ICSF setup required?


OnnoVdT:
## Description

1. For starters, this is not in ZWESECUR, so it belongs somewhere else. 
2. Where does Zowe use symmetric keys? As far as I know, or North and South-bound communication is all doen with asymetrical private/public keys.
 
## Pages to Update

https://docs.zowe.org/stable/user-guide/configure-zos-system.html#configure-an-icsf-cryptographic-services-environment

## Expected behavior

Move to correct location, and give more info on when this is needed.

nannanli:
@NayerNajafi @Joe-Winchester Hi Nayer, Joe, 

There was a discussion about this previously and Nayer commented: 

> Generating symmetric keys (Symmetric Encryption) is required if we want to have cryptography services such as Data Confidentiality to protect data from disclosure to unauthorized parties, so it is not a requirement for Zowe. I would say the part related to generating symmetric keys (defining RACF profiles) should be removed  from Zowe document. We could only have the steps about defining or checking the ICSF configuration depending on whether ICSF is already installed. 

I'm wondering if we should remove the https://docs.zowe.org/stable/user-guide/configure-zos-system.html#configure-an-icsf-cryptographic-services-environment section or alternatively, clearly state that the steps are optional.

Joe-Winchester:
@OnnoVdT - do you think we should do what @nannanli suggests and remove the section about generating ICSF ?  I'm in favor of removing it which is what @NayerNajafi is suggesting but I wanted to get your opinion as well before we pull the doc section.

OnnoVdT:
I agree with removing. If it stays it has to be in a different section that properly frames the context, and the optional nature.

nannanli:
@Joe-Winchester @OnnoVdT Thank you for confirming! I'll remove this section in the coming release.

1000TurquoisePogs:
I believe the description for why this was in the documentation was wrong, but it actually was a required step of running zss. 1.11 and 1.12 may be broken on new systems currently.

The real reason this section was in the documentation:
ZSS cookies require random number generation for security. ICSF is a secure way to do that, so https://github.com/zowe/zowe-common-c/blob/master/c/icsf.c#L533
CSNBRNGL is invoked.

https://www.ibm.com/support/knowledgecenter/en/SSLTBW_2.3.0/com.ibm.zos.v2r3.csfb400/rng.htm

[...]

PR #1142, opened 2020-04-22
Remove ICSF config section

nannanli:
#991

Issue number: #991


Joe-Winchester:
@ifakhrutdinov - we are removing the section telling users how to configure ICSF from the Zowe docs.  
In https://docs.zowe.org/stable/user-guide/configure-xmem-server.html#step-5-configure-an-ivsf-cryptographic-services-environment this references this section of the docs.
Is ICSF needed for default Zowe x-mem operation ?   If not then and only needed for extenders (which I believe is the case) then we might move this section elsewhere.

ifakhrutdinov:
@Joe-Winchester no, we don't need ICSF config for xmem. What is the reason behind these changes? Does the default config work for most people?

Joe-Winchester:
> @Joe-Winchester no, we don't need ICSF config for xmem. What is the reason behind these changes? Does the default config work for most people?

We are trying to doc just the minimum config pre-req required for customers, and we have had feedback that ICSF configuration is difficult and questioning as to why we call it out, so for 1.11 we're updating the docs to just remove all references of it from requirements, chapters saying how to do it, ...

nannanli:
@Joe-Winchester Other references to ICSF have been removed. See https://github.com/zowe/docs-site/pull/1142/files. 
FYI - @NayerNajafi

nannanli:
@Joe-Winchester @NayerNajafi Do you think there is a need to create a Medium blog to explain the details about ICSF for users who might need it?

Joe-Winchester:
> @Joe-Winchester @NayerNajafi Do you think there is a need to create a Medium blog to explain the details about ICSF for users who might want to install it?

@nannanli - yes this is a good idea.  Do you want to create a git issue for the blog and we can schedule it for a sprint and this will be useful for customers.

nannanli:
@Joe-Winchester An issue was created https://github.com/zowe/docs-site/issues/1148. Also put together all previous information about ICSF in the description as a starting point and added you and @NayerNajafi there.

ifakhrutdinov:
@Joe-Winchester I think at a minimum we should have kept `The user who runs ZOWESVR will need READ access to CSFRNGL in the CSFSERV class.` XMEM doesn't need that but ZSS (non authorized http server) does need that, otherwise the desktop won't work.

PR #1171, opened 2020-05-01
Semantic versioning

Joe-Winchester:
#### Semantic versioning
Zowe uses semantic versioning for releases so this doc update is to switch from calling Zowe release v.r.m to be V.v.p to align with Semver, and also to update the troubleshooting to let users know how to determine what Zowe release they have (which came from a customer situation where they asked on slack)

Issue number: #1172


docs/troubleshoot/troubleshooting.md
```diff
@@ -9,3 +9,34 @@ To isolate and resolve Zowe&trade; problems, you can use the troubleshooting and
...
 - [Troubleshooting Zowe using Diagnostics Data Script](troubleshoot-diagnostics.md)
+
+## Zowe releases
+
+Zowe uses semantic versioning for its releases, also known as Semver.  Each release has a unique ID made up of three numbers separted by periods.  
+
+```
+Major Version. Minor Version. Patch.
+```
+Each time a newer release is created the release ID is incremented.  Each number represents the content change since the previous release.  For example, `1.5.0` represents the firth minor release since the first major release.  `1.5.1` represents the first patch to the `1.5.0` release, and `1.6.0` is the first minor release to be created after `1.5.1`.
```
solsu01, line 20: firth - typo

docs/troubleshoot/troubleshooting.md
```diff
@@ -9,3 +9,34 @@ To isolate and resolve Zowe&trade; problems, you can use the troubleshooting and
...
+
+### Patch
+A patch is usually reserved for a bug fix to a minor release.
+
+### Minor release
+A minor release indicates that functionality has been added but the code is backwards compatible.  The Zowe community work on two week sprints and creates a minor release at the end of these, typically once per month although the frequency may vary.
+
+### Major release
+A major release indicates that changes have been made to the public API and the code is no longer backwards compatible.  This is sometimes refered to as a "breaking" change.  Currently Zowe is at version 1.  This is associated with the Zowe conformance program, see [Zowe conformance program](../extend/zowe-conformance-program.md), such that offerings that extend Zowe and achieve the Zowe conformance badge will remain compatible with Zowe throughout its version 1 lifetime.
```
solsu01, line 29: Do you think we would do major releases in some cases **without** the introduction of breaking changes? Are there other reasons that would lead to a major release?

I realize semver says major release is directly related to breaking API changes, but I'm wondering if we would hold ourselves to that.

Issue #1180, opened 2020-05-04
Document validation mode of 'running' Zowe


stevenhorsman:
Document changes for https://github.com/zowe/zowe-install-packaging/issues/1239

PR #1181, opened 2020-05-04
[Post-v1.11] Add validate mode info

stevenhorsman:
To be merged after https://github.com/zowe/zowe-install-packaging/pull/1335

Signed-off-by: stevenhorsman <steven@uk.ibm.com>

Issue number: #1180

### DIFF
--- a/docs/extend/lifecycling-with-zwesvstc.md
+++ b/docs/extend/lifecycling-with-zwesvstc.md
@@ -29,7 +29,7 @@ If present, the `validate.sh` script performs tasks such as:
 
 During execution of the `validate.sh` script, if an error is detected, then a component should echo a message that contains information to assist a user diagnosing the problem.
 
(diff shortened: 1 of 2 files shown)


stevenhorsman:
@BrandonJenkins14 - thanks for making sure - but this is doc-ing new function which isn't in 1.11, so deliberately against doc-staging

nannanli:
@stevenhorsman Hi Steve, I tweaked the wording to be as follows, which might sound more task-oriented and from the user's perspective. Let me know if there is any technical inaccuracy. Thanks! 

> If you are a Zowe administrator, you might want to check whether all the component validation checks of the Zowe installation pass without starting any of the components. To do this, you can add `VALIDATE_ONLY=true` to the `instance.env` file. Then, Zowe will not be launched after the validation stage.

nannanli:
zowe/zowe-install-packaging#1335 has been merged to staging, so merging the docs in this PR.

Issue #1164, opened 2020-04-29
[Troubleshooting tip] - node cluster timeout - spotted in slack


stevenhorsman:
### 1. Slack conversation link (if any)
https://openmainframeproject.slack.com/archives/CC08782AG/p1588083435121600

### 2. Choose a title
Node cluster storage timeout

### 3. Symptom
When starting the Zowe server the node address spaces don't start

### 4. Solution
edit $INSTANCE_DIR/instance.env to add ZWED_node_cluster_storageTimeout=30000 or some other number.

1000TurquoisePogs:
In 1.11, the default has been set to 30000 (was 5000). That SHOULD be good for everyone, but in the case that it fails, this number can be set higher. It's in milliseconds.

PR #1182, opened 2020-05-04
Zds 1164

Joe-Winchester:
#1164

### DIFF
--- a/docs/troubleshoot/app-framework/app-troubleshoot.md
+++ b/docs/troubleshoot/app-framework/app-troubleshoot.md
@@ -192,4 +192,48 @@ port: 7553 }
 **Solution:**   
 You can ignore these messages. These messages are timing-related where different Eureka servers come up, try to connect to each other, and warn that the endpoint they are trying to perform a handshake with is not available.  When all of the Eurka services have started, these errors will stop being logged.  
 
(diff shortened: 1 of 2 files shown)


Joe-Winchester:
Tls @stevenhorsman - I've just told users to upgrade to 1.11 for now, which means at least if they see the error with pre 1.11 they'll be able to resolve it

NakulManchanda:
I think we can add suggestion as pointed by steve as well to docs

$INSTANCE_DIR/instance.env
ZWED_node_cluster_storageTimeout=30000

as discussed on architecture call today is not a bad idea.

Joe-Winchester:
> we can suggest user to add this variable,
> $INSTANCE_DIR/instance.env
> ZWED_https_node_cluster_storageTimeout=30000
> as discussed today, i can verify the exact variable name

@NakulManchanda - I wasn't sure from looking at the code and also the issue whether the ZWED_http:node_cluster_storageTimeout=30000 actually is implemented and functional, or whether it was a desire to have ZLUX enhanced to do this.    If it is functional, how many other similar ZWED_ variables are available that might be interesting to doc ?

stevenhorsman:
I decided to cut out the middle man and just pushed the fix myself

PR #1191, opened 2020-05-05
Updates to 1.11 summary of changes

Joe-Winchester:
The 1.11 release notes didn't include information on the changes made to the Zowe installer.
I noticed it doing a quick look so this is to add into master info on what was included in 1.11 relating to the installer and related doc updates.

Because this is for updating 1.11 which is live I've set the branch to be /master, but I'm not sure how this then gets into docs.zowe.org so maybe @nannanli or @jackjia-ibm knows how to do that once the PR is approved.

### DIFF
--- a/docs/getting-started/summaryofchanges.md
+++ b/docs/getting-started/summaryofchanges.md
@@ -96,21 +96,22 @@ Review the [Zowe Explorer Change Log](https://github.com/zowe/vscode-extension-f
 
 You can install the latest version of the extension from the [Visual Studio Code Marketplace](https://marketplace.visualstudio.com/items?itemName=Zowe.vscode-extension-for-zowe).
 
-### Zowe installer
+#### Zowe installer
 
-- A `-l` parameter that controls where the setup scripts write trace logs is added to [zowe-setup-certificates.sh](../user-guide/configure-certificates.md#generate-certificate-with-the-custom-values), [zowe-install-xmem.sh](../user-guide/configure-xmem-server.md#copy-cross-memory-data-set-members-automatically) and [zowe-install-proc.sh](../user-guide/configure-zowe-server.md#step-1-copy-the-proclib-member-zwesvstc)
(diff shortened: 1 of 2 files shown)


docs/getting-started/summaryofchanges.md
```diff
@@ -96,6 +96,22 @@ Review the [Zowe Explorer Change Log](https://github.com/zowe/vscode-extension-f
 
 You can install the latest version of the extension from the [Visual Studio Code Marketplace](https://marketplace.visualstudio.com/items?itemName=Zowe.vscode-extension-for-zowe).
 
+### Zowe installer
+
+- A `-l` parameter that controls where the setup scripts write trace logs is added to `zowe-install.sh`, `zowe-configure-instance.sh`, `zowe-create-keystore.sh`, `zowe-install-xmem.sh` and `zowe-install-proc.sh`.  The problem determination gathering script `zowe-support.sh` has a `-l` parameter to specify the locaiton of where the scripts have been collected.
```
stevenhorsman, line 101: `-l` is not on zowe-configure-instance.sh. There is no such thing as `zowe-create-keystore.sh` - I assume you mean `zowe-setup-certificates.sh`? zowe-support.sh's -l paramter didn't make it into 1.11

Joe-Winchester: My bad @stevenhorsman .  I've fixed this and also removed the `zowe-install.sh` reference as I think that was in 1.10 already, and I've added doc links to the relevent sections to where each script is described

stevenhorsman: `-l` wasn't in `zowe-install.sh` in 1.10, so it should be added back in!

nannanli:
@Joe-Winchester @stevenhorsman I made some edits to the release notes. This PR contains information about the ZAAS client which I guess should not be included in this PR. It might be because this PR was initially created from the docs-staging branch. I'll close this PR and open a new one.

Issue #1210, opened 2020-05-13
More detail on CLI on z/OS as customers are running into questions about support


Joe-Winchester:
# Description
Customers trying to install CLI on z/OS are running into problems where they can't get secure credentials to work nor can they get the IBM DB2 CLI)

## Pages to Update
https://docs.zowe.org/stable/user-guide/systemrequirements.html#zowe-cli-requirements-client

# Expected behavior
Document that
```
The IBM DB2 and Secure Credential Store plug-ins for Zowe CLI will not run on z/OS due to native code requirements.
```

dkelosky:
As a side note, we'll soon have a login command that will store JWTs instead of plain user names and passwords.

These should be kept secret, but are probably a step up on security over plain text user and password.

Joe-Winchester:
Tks @dkelosky - I created a PR to the system requirements chapter as quite a few users were running into issues and needing 1:1 help and clarification on slack https://github.com/zowe/docs-site/pull/1211.  Feel free to review if you have time.
Looking forward very much to seeing the CLI SSO implementation - is there a git issue I can see how that's progressing or maybe a topic for an architecture call to do a presentation, or is there a CLI scrum I (and others) could join who tend to work on z/OS side but are keen to see CLI and z/OS integration occurring

dkelosky:
Hi Joe, I added one comment to your PR.  I created https://github.com/zowe/zowe-cli/issues/718 for a brief intro into work that we're doing now.  More details of our plans can be found [here](https://medium.com/zowe/zowe-cli-token-authentication-mfa-and-sso-b88bca3efa35).  

Would you like to talk this over on the Zowe arch call?

Joe-Winchester:
@dkelosky - yes I think for next week's arch 5/19/2020 call it'd be great if you could share the design and thinking of the Zowe CLI SSO.  cc: @1000TurquoisePogs

dkelosky:
OK I'll plan for that.  I probably don't have a lot of diagrams but will be able to use the blog to help with discussion.

nannanli:
PR #1211 has been merged. Closing this issue but feel free to reopen if any problem persists.

PR #1211, opened 2020-05-13
system requirements changed

Joe-Winchester:
### Is there a related issue for this PR? 
Issue number: #1210 

Updated the system requirements to make it clearer that CLI on z/OS doesn't support DB2 and Secure Credentials plugin


docs/user-guide/systemrequirements.md
```diff
@@ -115,7 +125,7 @@ Configure zowe-setup-certifcates.env using the following parameters. Both are re
 
 ## Zowe CLI requirements (client)
 
-Zowe CLI is supported on platforms where Node.js 8.0 or 10 is available, including Windows, Linux, and Mac operating systems. Zowe CLI was also tested and confirmed to run on Unix System Services (USS) on z/OS.
+Zowe CLI is supported on platforms where Node.js 8.0 or 10 is available, including Windows, Linux, and Mac operating systems.
```
dkelosky, line 128: Perhaps this should be "Node.js 8.0 or higher" LTS versions (as is stated below).

PR #1241, opened 2020-06-01
Update presentation to fix alignment

stevenhorsman:
Update presentation to fix alignment


nannanli:
@stevenhorsman I updated the toc to match the current order. Also updated the screen capture based on the diagram. Others all look good to me. Thanks!

PR #1262, opened 2020-06-09
Add note about PROGCTL before APPLY

John-A-Davies:
Fix for https://github.com/zowe/zowe-install-packaging/issues/1428

https://github.com/zowe/zowe-install-packaging/issues/1428


John-A-Davies:
@stevenhorsman @Joe-Winchester I've added a doc note, but the PERMIT could be added as a prior step in the ZWE7APLY job - what do you think?

@OnnoVdT I think we need READ access to PROGCTL because `zssServer` has the `p` attribute.

PR #1266, opened 2020-06-11
Bpx.fileattr.progctl

John-A-Davies:
Issue number: #1262


OnnoVdT:
should we say read access to BPX.FILEATTR.PROGCTL`, if the profile exists` ?

PR #1268, opened 2020-06-11
Create config-cert-instance-hints.md

John-A-Davies:
This PR provides a couple of hints and tips about specifying hostname and IP address during configuration of certificates and an instance of Zowe.

Issue number: #1279


docs/user-guide/config-cert-instance-hints.md
```diff
@@ -0,0 +1,54 @@
+Hints about creating the certificate and instance when configuring Zowe.
+
+When you are configuring Zowe on z/OS, you need to create certificates, and then create the Zowe instance.
+
+### Creating certificates
+
+We recommend that you re-create the certificates after every Zowe upgrade.
```
stevenhorsman, line 7: We don't currently recommend this - keystore generation is listed as a one-time task

docs/user-guide/config-cert-instance-hints.md
```diff
@@ -0,0 +1,54 @@
+Hints about creating the certificate and instance when configuring Zowe.
```
stevenhorsman, line 1: Why is this information not going in the existing doc pages for setting up certificates and instances?

John-A-Davies: It's here as a placeholder - a basis for discussion.  I hope it will be edited into the relevant section of the doc in due course.

docs/user-guide/config-cert-instance-hints.md
John-A-Davies: Ok, then I misunderstood.  Is there one keystore per instance then?

John-A-Davies: The new HOLDDATA says this - is this a conflict?
```
Run <ROOT_DIR>/bin/zowe-setup-certificates.sh again with an 
updated zowe-setup-certificate.env to pick up the new 
definitions.
```

stevenhorsman: 1.12 has some extra parameters - hence is in the HOLDDATA, but it isn't mandatory, or usual

stevenhorsman: > Ok, then I misunderstood. Is there one keystore per instance then?

No

John-A-Davies: So, 
1.  normally, you run `zowe-setup-certificates.sh` only the first time you install Zowe, unless instructed otherwise by HOLDDATA or the release notes for that release.
2.  the keystore is shared by all instances.

stevenhorsman: 1 - yes
2 - it doesn't have to be, but that is the way Joe has doc'd it and our default recommendation if you create the keystore directory in /global/zowe/keystore

PR #1284, opened 2020-06-23
Add new file for the api gateway conf params

taban03:
Signed-off-by: at670475 <andrea.tabone@broadcom.com>

### Description 
Add new markdown to collect GW parameters which can be set during Zowe runtime


balhar-jakub:
Shouldn't the new file be in the user-guide?

This doesn't really feel like that strongly linked to the extending of the API ML.

taban03:
> Shouldn't the new file be in the user-guide?
> 
> This doesn't really feel like that strongly linked to the extending of the API ML.

Yes, I think you're right. I will move it

PR #1293, opened 2020-07-01
prefer IP address option for enablers

achmelo:
Signed-off-by: achmelo <a.chmelo@gmail.com>

Option to prefer IP address at registration to eureka for enablers.

### DIFF
--- a/docs/extend/extend-apiml/onboard-plain-java-enabler.md
+++ b/docs/extend/extend-apiml/onboard-plain-java-enabler.md
@@ -180,6 +180,7 @@ are written in `${parameterValue}` format. For your service configuration file, 
  description: Sample API ML REST Service
  baseUrl: https://${samplehost}:${sampleport}/${sampleservice}
  serviceIpAddress: ${sampleHostIpAddress}
+ preferIpAddress: false
 
  homePageRelativeUrl: /application/home
  statusPageRelativeUrl: /application/info
@@ -295,10 +296,12 @@ The onboarding configuration parameters are broken down into the following group
 
 *  **serviceIpAddress** (Optional)
 
-
     This parameter specifies the IP address of the service and can be provided by system administrator in externalized service configuration.
     If this parameter is not present in the configuration file or is not set as a service context parameter, it will be resolved from the hostname part of the `baseUrl`.
 
+* **preferIpAddress** (Optional)
+
+    In some cases, it is preferable for Eureka to advertise the IP addresses of services rather than the hostname. Set preferIpAddress to `true` and, when the application registers with eureka, it uses its IP address rather than its hostname.
 
 ### Administrative endpoints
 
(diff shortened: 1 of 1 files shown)


docs/extend/extend-apiml/onboard-plain-java-enabler.md
```diff
@@ -295,10 +296,12 @@ The onboarding configuration parameters are broken down into the following group
 
 *  **serviceIpAddress** (Optional)
 
-
     This parameter specifies the IP address of the service and can be provided by system administrator in externalized service configuration.
```
IgorCATech, line 299: "Specifies the service IP address and can be provided by a system administrator in the externalized service configuration.

docs/extend/extend-apiml/onboard-plain-java-enabler.md
```diff
@@ -295,10 +296,12 @@ The onboarding configuration parameters are broken down into the following group
...
     If this parameter is not present in the configuration file or is not set as a service context parameter, it will be resolved from the hostname part of the `baseUrl`.
 
+* **preferIpAddress** (Optional)
+
+    In some cases, it is preferable for Eureka to advertise the IP addresses of services rather than the hostname. Set preferIpAddress to `true` and, when the application registers with eureka, it uses its IP address rather than its hostname.
```
IgorCATech, line 304: The first sentence sounds a bit vague. Can we specify the cases when it's preferable to advertise IP addresses instead of hostnames? 
If my understanding is correct, this parameter "specifies a service IP address instead of hostname if the value is "true"
The second sentence might sound like this:
"If the parameter is set to "true", an application that is registered with Eureka will use its IP address instead of hostname." but it's redundant if my first suggestion is correct. If so, I would specify the cases in which it's preferable to advertise an IP address instead of hostname, assuming this information will help our customers. If it's not helpful but confusing, I would not mention it at all.

achmelo: I think that anyone can have different case for using it, maybe someone is using certificate issued for IP address or hostname cant be resolved, but there might be others. So it would be better just inform that there is this option.

IgorCATech: Ok, then. To be consistent with the way descriptions are formulated, the first sentence will sound as follows: 
"Specifies an IP address of an application registered with Eureka if the value is "true". 
The second sentence is "Use the parameter if you want to advertise a service IP address instead of its hostname."

achmelo: Second part is good. For the first, you dont need to specify ip address, it will be resolved from hostname. So could it be something like: "Set this parameter "true" if you want to advertise a service IP address instead of its hostname. "

IgorCATech: Yes, that sounds good. Here's a little change to your piece "Set the value of the parameter to "true" if you want to advertise a service IP address instead of its hostname."

Issue #1399, opened 2020-09-08
Zowe icon in the tab for the docs site


zdmullen:
I would very much like to see the Zowe icon in the tab for an open page of the Zowe Docs Site. 

## Is your request for enhancement related to a problem? Please describe.
Currently, I see only the drab image of the world:
![image](https://user-images.githubusercontent.com/38981509/92455888-85440b80-f1c2-11ea-8113-8e1758256450.png)

## Describe the solution you'd like
Could we add in the Zowe icon?

BrandonJenkins14:
T

PR #1407, opened 2020-09-10
Fix favicon

BrandonJenkins14:
The favicon does not appear live site. Added a link from head so that it can publish the icon.  Addresses #1399


Roavill2 (approved):
Approved provided it passes tests

PR #1412, opened 2020-09-14
Zowe Explorer Associate Profiles feature (doc upd)

IgorCATech:

Adding a description of the associate profile feature in Zowe Explorer (VS Code extension). 
The issue opened in the zowe-explorer repo https://github.com/zowe/vscode-extension-for-zowe/issues/972

:heart:Thank you!


IgorCATech:
Hey @BrandonJenkins14 Thanks for your feedback. Let me review it a bit later, and I'll get back to you ;)

docs/user-guide/ze-profiles.md
```diff
@@ -29,6 +29,25 @@ If you need to edit a profile, click the **Update Profile** button next to the c
 
 In addition, you can hide a profile from the explorer tree, and permanently delete a profile. When you delete your profile permanently, the extension erases the profile from the `.zowe` folder. To hide or delete a profile, right-click the profile and choose one of the respective options from the list.
 
+### Associate Profile
+
+Zowe Explorer allows you to broaden its functionality, using Zowe Explorer Extender APIs. Extenders add new functionality to the extension, leveraging a simple mechanism that enables you to associate profiles. Every Zowe Explorer profile has two types of association: primary and secondary. While the primary association is `zosmf`, the secondary association includes Zowe CLI plug-ins or other services you might have. Within the scope of their own extender package, associated profiles can be used to access a relevant REST API that is available to the extender.
```
zdmullen, line 34: I think these first two lines are far too general and off the topic at hand. 

Perhaps it would be better to say something along the lines of "Extensions built on Zowe Explorer provide users with additional functionality through unique profile types, such as RSE or FTP. The "associate profiles" function allows users to link, or associate, these extension profile types with an existing zOSMF profile. Every Zowe Explorer profile has two types of association..." Then alter the following sentences to follow this idea...

docs/user-guide/ze-profiles.md
```diff
@@ -29,6 +29,25 @@ If you need to edit a profile, click the **Update Profile** button next to the c
...
+**Follow these steps:**
+
+1. Navigate to the explorer tree.
+2. Click the associate profiles button.
+3. Select a secondary profile type.
+4. Select a file name you want to create an association for.
+
+<img src="../images/ze/ZE-profile-links.gif" width="600" height="300" alt="Edit a Profile">
+
+You have successfully associated your Zowe Explorer profile.
```
zdmullen, line 47: with the specified secondary or extension profile type, perhaps?

docs/user-guide/ze-profiles.md
```diff
@@ -67,3 +86,27 @@ Activate the SCS plug-in in Zowe Explorer.
 6. Restart VSCode.
 
 The credentials of your newly created or existing profiles are now stored securely.
+
+## Use Base Profile and Token with Existing Profiles
+
+As a Zowe user, you can leverage the base profile functionality to access multiple Zowe services instantly. Base profiles enable you to authenticate via Zowe API Mediation Layer. You can use base profiles with more than one service profile. Any service profile with the same host and port as your base profile automatically uses your API ML token to log in.
+
+Before using the base profile functionality, ensure that you have [Zowe CLI](https://docs.zowe.org/stable/user-guide/cli-installcli.html) v1.13.0 or higher installed.
```
jellypuno, line 94: it should be Zowe CLI v6.16

crawr:
Not sure if statement below is worth repeating here, since it is written in the Zowe CLI doc. Zowe Explorer will issue an error message if token is expired, but it will be the same error message as with regular Invalid Credentials.

>Tokens expire after a period of time defined by your security administrator. When a token expires, you must log in again to get a new token.

docs/user-guide/ze-profiles.md
```diff
@@ -67,3 +86,27 @@ Activate the SCS plug-in in Zowe Explorer.
 6. Restart VSCode.
 
 The credentials of your newly created or existing profiles are now stored securely.
+
+## Use Base Profile and Token with Existing Profiles
+
+As a Zowe user, you can leverage the base profile functionality to access multiple Zowe services instantly. Base profiles enable you to authenticate via Zowe API Mediation Layer. You can use base profiles with more than one service profile. Any service profile with the same host and port as your base profile automatically uses your API ML token to log in.
```
zdmullen, line 92: I don't know if we want to say "Zowe services". I think it can just be "services," as is described in this section: https://docs.zowe.org/stable/user-guide/cli-usingcli.html#using-profiles

Actually, if you can re-create much of the information from the above link, I think that would be helpful.

docs/user-guide/ze-profiles.md
```diff
@@ -67,3 +86,27 @@ Activate the SCS plug-in in Zowe Explorer.
...
+
+Before using the base profile functionality, ensure that you have [Zowe CLI](https://docs.zowe.org/stable/user-guide/cli-installcli.html) v6.16 or higher installed.
+
+Connect your service profile with a base profile and token.
```
zdmullen, line 96: Should this be a heading? ### ?

zdmullen: And maybe "Log in and connect your service profile..."

docs/user-guide/ze-profiles.md
```diff
@@ -67,3 +86,27 @@ Activate the SCS plug-in in Zowe Explorer.
...
+
+   A local base profile is created that contains your token. For more information about the process, see [Token Management](https://docs.zowe.org/stable/user-guide/cli-usingcli.html#how-token-management-works).
+  
+3. Run Zowe Explorer and click the **+** icon.  
+
+4. Select the profile you use with your base profile with the token.
+
+   The profile appears in the tree and you can now use this profile to access z/OSMF via the API Mediation Layer.
+
+For more information, see [Integrating with API Mediation Layer](https://docs.zowe.org/stable/user-guide/cli-usingcli.html#integrating-with-api-mediation-layer).
```
zdmullen, line 112: Good otherwise. We'll expand on this more in the near future, believe you me 👍

docs/user-guide/ze-profiles.md
katelynienaber: With removing "Zowe" I think it reads fine. We can add more as we see confused questions on Slack ❓ ❔

docs/user-guide/ze-profiles.md
katelynienaber: Maybe you could say "How to connect..." and make it a heading, otherwise it is sort of confusing why this sentence is here P:

docs/user-guide/ze-profiles.md
katelynienaber: I would maybe add one note that says something about, if you have an existing profile that you want to use with the new base profile, check that the service profile is using the same host & port as the base profile.

katelynienaber (changes requested):
Looks great! Only the last comment is a change I think is really necessary, since people might be confused how to use existing profiles with the new base profile.

Hopefully it's pretty straightforward, and people aren't too confused ^^'

jellypuno:
> I would maybe add one note that says something about, if you have an existing profile that you want to use with the new base profile, check that the service profile is using the same host & port as the base profile.

I think you don't have to add that note. When a user has a base profile, he/she should know what is the use of it.

docs/user-guide/ze-profiles.md
IgorCATech: I guess by services I meant components/applications. If I replace "services" with any of those, will it be clearer?

docs/user-guide/ze-profiles.md
IgorCATech: I didn't want to entitle such a small paragraph, really. Does it rather confuse you?

docs/user-guide/ze-profiles.md
IgorCATech: @katelynienaber I moved this suggestion of yours to line 92. I didn't just cut it out :)

IgorCATech:
@jellypuno @katelynienaber re the last Katelyn's comment, the note has been added to the meta-description (the first paragraph of the section)

docs/user-guide/ze-profiles.md
jellypuno: > Any service profile with the same host and port as your base profile automatically uses your API ML token to log in.

This is not true. I can put my host and port and username and password in my service profile and it will not use the token from my base profile.

docs/user-guide/ze-profiles.md
```diff
@@ -91,7 +91,7 @@ The credentials of your newly created or existing profiles are now stored secure
 
 As a Zowe user, you can leverage the base profile functionality to access multiple Zowe services instantly. Base profiles enable you to authenticate via Zowe API Mediation Layer. You can use base profiles with more than one service profile. Any service profile with the same host and port as your base profile automatically uses your API ML token to log in.
 
-Before using the base profile functionality, ensure that you have [Zowe CLI](https://docs.zowe.org/stable/user-guide/cli-installcli.html) v6.16 or higher installed.
+Before using the log in and connect your service profile, ensure that you have [Zowe CLI](https://docs.zowe.org/stable/user-guide/cli-installcli.html) v6.16 or higher installed.
```
zdmullen, line 94: Before you log in and connect your service profile...?

zdmullen:
And the previous comments still need to be addressed.

docs/user-guide/ze-profiles.md
BrandonJenkins14: Some thoughts that might help (I worked on the CLI profile doc):
- For your first comment about services, I agree to omit the word "Zowe". Also keep in mind that you can only do SSO through APIML **if** the service has been configured/registered to the layer. For services that are not registered to the layer, you still need to supply host, port, etc... explicitly on commands or service profiles. 
- The benefit isn't that you can access services more "instantly". It's that you can point your commands at APIML to reap the benefits that API ML provides, such as SSO to multiple services, load balancing, security, etc... 
- I agree with @zdmullen comment that you might want to cherry-pick some of the CLI profiles doc and edit it as needed for Explorer doc. There are several important things described there that you might not want to write about from scratch, like base vs service profile, how the `--base-path` option works, how to do SSO but still use your service profiles to connect to regular services that aren't registered to APIML, etc.... Basically translate that doc to one about doing the same things using Explorer GUI.

BrandonJenkins14: @jellypuno Agreed. It's more like "if you don't supply a host and port in your service profile, then the tool references your base profile and looks for a token to use instead. Omit the host and port from service profiles to "instruct it" to connect to APIML, along with providing --base-path of the APIML instance.

docs/user-guide/ze-profiles.md
```diff
@@ -67,3 +86,27 @@ Activate the SCS plug-in in Zowe Explorer.
...
+
+1. Open Zowe CLI and run the following command: `zowe auth login apiml`.
+
+2. Follow the instructions to complete the login.
+
+   A local base profile is created that contains your token. For more information about the process, see [Token Management](https://docs.zowe.org/stable/user-guide/cli-usingcli.html#how-token-management-works).
+  
+3. Run Zowe Explorer and click the **+** icon.  
+
+4. Select the profile you use with your base profile with the token.
```
BrandonJenkins14, line 108: I'm not familiar with the functionality in Explorer, but is there a point in this procedure where the user needs to remove host/port from that service profile, or else their command won't get routed through API ML? Should we remind them to do that here? 

Maybe this? "Select the profile that you want to use to access a particular mainframe service. Ensure that the profile does not contain host or port. When you omit host and port, Explorer uses the token in your base profile to connect to API ML and route the command to an appropriate instance of the service. 

Is there a `--base-path` concept in Explorer? Should it be mentioned?

docs/user-guide/ze-profiles.md
BrandonJenkins14: Opinion: Rename the existing heading to something catchy that a user wants to do, like "Access services through API ML with SSO" or similar.
And move that rogue sentence into the above paragraph so that it doesn't stick out. The paragraph explains the concept and then "in order to make this work, you need to connect your service profile with a base profile. "

docs/user-guide/ze-profiles.md
IgorCATech: @BrandonJenkins14 To your first point, wrt to accessing API ML, is this information (that the service has to be registered to layer to do SSO) available on the CLI/API ML page? I think I should link to it.
point 2 - right, thanks for clarifying this to me. I should change that part accordingly. 
last point - I guess I should do more research. This feature is more than meets the eye.

jellypuno:
@IgorCATech @BrandonJenkins14 @zdmullen It is important to keep in mind that the base profile functionality in Zowe Explorer is not the same as in Zowe CLI. We implemented the most basic feature in explorer which is to **load** the zosmf profile that uses base profile. Are assumption is that the users are knowledgeable in this area.

Any manipulation in either profile, whether it is a create or update, will be done in Zowe CLI. So I would suggest that we keep it simple and we just reference the users to Zowe CLI documentation.

docs/user-guide/ze-profiles.md
IgorCATech: @katelynienaber Would you be able to answer the first 2 Brandon's questions? Also, the question about the --base-path concept.

docs/user-guide/ze-profiles.md
BrandonJenkins14: @IgorCATech 

- https://docs.zowe.org/stable/extend/extend-apiml/onboard-overview.html  (an API developer or admin of some sort would likely do this) 
- check out the more recent comments, it seems that I went overboard with my suggestion to make significant changes. Might be close to complete here :P

IgorCATech:
@BrandonJenkins14 It's a good suggestion but I wouldn't add this just yet. Maybe, we should roll out this doc and see how customers react. If we receive feedback saying that the doc lack this info, we'll add it then. @zdmullen What do you think?

zdmullen:
> @BrandonJenkins14 It's a good suggestion but I wouldn't add this just yet. Maybe, we should roll out this doc and see how customers react. If we receive feedback saying that the doc lack this info, we'll add it then. @zdmullen What do you think?

As long as we have the link to the CLI doc there, which we do, I would call it good for now. We'll have to change this in the near future anyway, with the PR that Richelle is eventually going to open. So let's go with it for now. Thanks.

katelynienaber (approved):
Lgtm! Thanks for all the changes @IgorCATech

katelynienaber (approved):
Lgtm! Thanks for all the fixes :D

docs/user-guide/ze-usage.md
```diff
@@ -148,6 +149,15 @@ Your PDS member (or PS) is uploaded.
 
 <img src="../images/ze/ZE-mult-profiles.gif" alt="Add Profile">
 
+#### Allocate Like
+
```
zdmullen, line 153: So maybe just a description below the title like:

"Select a data set and create a copy with the same parameters."

IgorCATech: That's the whole idea - we keep things short. I think developers appreciate this way more. Maybe, @jellypuno can speak to it

jellypuno: Or Duplicate a dataset?

IgorCATech: Also, I do include the description of the feature. Line 44

PR #1435, opened 2020-09-29
Refactor note

janan07:
This PR is to rearticulate the note from:

Currently, API ML can only provide SSO in a domain with a single method of security in place.
to
Currently, API ML can provide SSO only in a single security domain.

This improves the accuracy of the SSO domain limitations


pz891284:
Agree.

pz891284 (approved):

BrandonJenkins14 (approved):

PR #1476, opened 2020-11-06
Hold off client certificate auth documentation for this release

jandadav:
This is documentation revert for client cert auth, which should not go to the upcomming release. It will be added back when the client cert auth is ready.

There is no need for TI review imho.


nannanli (approved):
LGTM - Thanks for the updates!

PR #1502, opened 2020-12-04
Consistently describe npm requirement for CLI

BrandonJenkins14:
Information about npm prerequisite was outdated in the Install CLI topic. Users now require a currently-supported LTS version of npm. 

Copied the more current information from CLI Software Requirements so that the information is correct and consistent across the doc.


zFernand0 (approved):
LGTM!
Wondering what you think about the duplication of steps 1 and 2 in both methods. 😋

PR #1513, opened 2020-12-10
Create install-cli-via-proxy.md

MichaelABroadcom:

### DIFF
--- a/docs/user-guide/install-cli-via-proxy.md
+++ b/docs/user-guide/install-cli-via-proxy.md
@@ -1,0 +1,70 @@
+​
+# Install CLI from Online Registry Via Proxy
+​This topic describes how to install Zowe CLI using the NPM install command when you are working behind a proxy server. You will need to use this installation method if your site blocks access to public NPM.
+​​
+You can install Zowe CLI from an online registry via proxy on Windows, macOS, or Linux operating systems:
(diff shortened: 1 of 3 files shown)


BrandonJenkins14 (approved):

zFernand0 (approved):
LGTM!
Thanks!

docs/user-guide/install-cli-via-proxy.md
```diff
@@ -0,0 +1,72 @@
...
+*  Your default registry must be public npm (or a mirror of public npm).
+*  If you previously installed the CLI and want to update to a current version, see Update Zowe CLI.
+
+​Follow these steps: 
+​
+1. Identify the proxy server, including the IP address or hostname and the port number.​
+    * If your proxy server **does not** require login credentials, issue the following commands to add the proxy URL to the NPM config file:
+
+        ```
+        npm config set https-proxy http://proxy.[proxy_name].com:8080
```
zFernand0, line 17: I believe the IP/hostname and port is the user identified is what will be plugged in here.
IDK if we need to change the variable `proxy_name` to identify that they need `http://[proxy_host]:[proxy_port]`

MichaelABroadcom: Thanks, good catch. I'll make that change.

MichaelABroadcom: I updated the text if you can check it.

PR #1708, opened 2021-06-07
Add .nojekyll file so that filenames starting with "_" get published

t1m0thyj:
The SDK typedoc added in https://github.com/zowe/docs-site/pull/1667 cannot be browsed online because clicking any link on [the main page](https://docs.zowe.org/stable/typedoc/index.html) triggers a 404 error. This worked when testing locally, but fails online because Jekyll/GitHub Pages removes all files that begin with an underscore (`_`).

In order to accommodate automatically generated HTML files whose names begin with an underscore, the suggested fix seems to be adding a `.nojekyll` file inside the "public" folder: https://github.com/vuejs/vitepress/issues/124#issuecomment-727819038

This PR implements the suggested fix. Since the issue isn't reproducible until the site is published, I'm unsure how to test this change before the PR is merged, and have opened the PR in draft mode.

PR #1713, opened 2021-06-09
Exclude docs/.vuepress/public subdirectories from link checker

t1m0thyj:
Follow up to #1708 which inadvertently broke [master build](https://wash.zowe.org:8443/blue/organizations/jenkins/docs-site/detail/master/315/pipeline) by making the link checker time out.

The directories inside docs/.vuepress/public (typedoc and web_help) contain automatically generated content that doesn't need to have its links checked, and makes the link checker take significantly longer.

This change should approximately halve the time of the "test" stage in the pipeline, since web_help folder used to be included in link check.

Thanks to @jackjia-ibm for helping me to fix this 🙂

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,8 @@
   "main": "index.js",
   "scripts": {
     "test": "echo \"Error: pending to verify dead links.\"",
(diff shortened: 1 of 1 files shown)


jackjia-ibm (approved):
Looks great! 9m test time.

Thank you Tim.

PR #1743, opened 2021-07-08
API mediation security - Tech Info language refactor

janan07:
This article has grown very long over the years and needs to be chunked into smaller articles to improve user experience. This PR is an initial language refactor. A subsequent PR will address the article chunking.


jandadav:
I've checked the edits and they look great to me! Thanks

jandadav (approved):

PR #1748, opened 2021-07-09
minor ha update

1000TurquoisePogs:
I just read through the existing HA documentation to familiarize myself with it and found just a few things to correct. Otherwise it makes sense to me! Thanks.


nannanli:
lgtm~ Will let Nayer @NayerNajafi take a look as well before merging.

nannanli (approved):

PR #1758, opened 2021-07-19
Add documentation about HA for app-server and zss

1000TurquoisePogs:
This updates the app framework documentation to explain how developers can make use of HA features, and what they can/cant do in an HA environment. Additionally I added some missing ZSS documentation while I was in that area of the doc.


lchudinov (approved):
Looks good to me.

PR #1891, opened 2021-11-09
User/joewinchester/keyringsupported

Joe-Winchester:
### Description (Keyring support) (including links to related git issues)

Customer in slack posted question about whether keyrings are supported as the doc still said that they were new in 1.15 and still experimental.  @jackjia-ibm replied that they are supported (thank you Jack), so this PR is to update the chapter.

Customer question on Oct 20th.   https://openmainframeproject.slack.com/archives/CC08782AG/p1634761526071100
```
Keystore versus key ring#
Has anyone successfully installed ZOWE using a z/OSMF shared keyring?
The doc is a little confusing.
The following doc says keyrings are not supported, but can be configured:
https://docs.zowe.org/stable/user-guide/configure-certificates/
The High Availability (ZOWE on a SYSPLEX) doc has KEYRING_NAME as requirement:
https://docs.zowe.org/stable/user-guide/install-ha-sysplex
```
Jack's reply 
```
Hi Michael, sorry about the doc, I think now keyring is fully supported for all core components. It’s possible to configure Zowe to use keyring https://docs.zowe.org/stable/user-guide/configure-certificates-keyring. Please let us know if there is problem. 
```

### Description (OMVSAPPL) READ needed

The Db2 Admin Foundation that is a Zowe Desktop extension had asked for an update to the Zowe pre-reqs to cover the need for SSO together with the troubleshooting tip that identifies when the step hasn't been done correctly.


jackjia-ibm (approved):
Looks great. Thanks Joe!

nannanli (approved):
Made some edits. LGTM

PR #2095, opened 2022-04-05
Add Wizard YAML prefill docs (v2)

CaioCamatta:
### Description

Adds documentation related to the new YAML prefill feature of the Wizard in the API layer. 

"As a user trying to onboard a service to API ML, I want to upload an existing partial or full configuration (YAML file) by using the Wizard UI." 

Related to PR 2201 in `api-layer`: https://github.com/zowe/api-layer/pull/2201
Related to Issue 2058 in `api-layer`: https://github.com/zowe/api-layer/issues/2058


docs/extend/extend-apiml/onboard-wizard.md
```diff
@@ -24,3 +24,8 @@ The following procedure describes how to onboard your REST service with the Wiza
    
 If you see your service in the list of API Catalog available services, you have onboarded your service successfully.
 
+### Prefilling fields
+
+After opening the wizard, you may upload a complete or partial YAML file to prefill the fields. The YAML file will be validated and the form fields will be populated.
+
+![YAML Prefill](../../images/api-mediation/wizard-yaml-prefill.png)
```
janan07, line 31: Please, trim grey margin from image.

janan07 (changes requested):
Please see comment Re: image edits

docs/extend/extend-apiml/onboard-wizard.md
CaioCamatta: Done

docs/extend/extend-apiml/onboard-wizard.md
```diff
@@ -24,3 +24,8 @@ The following procedure describes how to onboard your REST service with the Wiza
    
 If you see your service in the list of API Catalog available services, you have onboarded your service successfully.
 
+### Prefilling fields
+
+After opening the wizard, you may upload a complete or partial YAML file to prefill the fields. The YAML file will be validated and the form fields will be populated.
```
janan07, line 29: Could you please describe the procedure for uploading the YAML file? Something like this...
Follow these steps:
1.
2.
3.
You have successfully uploaded your YAML file.

janan07, line 29: Please change this statement as follows:
(Optional) Click *Choose File* to upload a complete or partial YAML file to prefill the fields. The YAML file is validated and the form fields are populated.

docs/extend/extend-apiml/onboard-wizard.md
janan07, line 31: I'm not clear how the option "Or fill the fields:" works. Do we need to provide any instructions to the user about which fields to fill and where, or is this clearly described in the Wizard?

CaioCamatta: That option just invites the user fill the form as they would normally, as in the screenshot attached. I don't think more instructions are necessary, but I would agree that the "Or fill the fields:" option looks a bit out of context in the image in the PR. So, I updated the image to exclude it.

image

docs/extend/extend-apiml/onboard-wizard.md
CaioCamatta: Updated.

docs/extend/extend-apiml/onboard-wizard.md
janan07: If I understand correctly, filling the fiends IS an option, but the feature you are trying to highlight is using the "Choose file" option, is that right? If so, it seems that the updated screenshot will suffice, but only if you don't want to bring the "fill the fields" option to the reader's attention. 

Also, I see this step appears after the validation step. Shouldn't this be before the validation step?

CaioCamatta: Correct, the user can either fill in the fields manually, or upload a pre-existing YAML file to populate some/all fields. So, in the screenshot we want to highlight the upload feature.

The YAML upload does happen before validation. I could put it as "Step 3 (Optional)", and move the current steps 3-6 up one number. What do you think?

CaioCamatta: @janan07 Done

janan07 (approved):
Reviewed and edited by Tech Info

PR #2107, opened 2022-04-07
Fix zss instance setup -> zis. Clarify AT-TLS use. Describe zss 64 bit

1000TurquoisePogs:
ZSS AT-TLS doc was a little silly because its not really zss specific, it applies to any server so I made it a little more generic by saying "appfw servers"

I added a section about zss 64 bit

I fixed that the section talking about setting up another ZSS was actually about ZIS all along. Very outdated!

### DIFF
--- a/docs/user-guide/mvd-configuration.md
+++ b/docs/user-guide/mvd-configuration.md
@@ -245,20 +245,50 @@ zss:
       attls: true
 ```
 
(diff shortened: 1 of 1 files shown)


1000TurquoisePogs:
Recreating this on @DivergentEuropeans branch

PR #2108, opened 2022-04-07
Server schema explanation

1000TurquoisePogs:
The logic that enforces this is not yet in the codebase, but it will be quite soon, so I'm putting this PR out as a line in the sand to describe how things will work, and I suggest we get this into the doc sooner than later as a disclaimer that it is coming.

### DIFF
--- a/docs/appendix/server-component-manifest.md
+++ b/docs/appendix/server-component-manifest.md
@@ -32,7 +32,7 @@ The manifest file contains the following properties:
 
   Defines the location of json schema files that are compatible with certain portions of Zowe as denoted by each child property.
   
-  * **`config`**
+  * **`configs`**
   
     Defines the location of the json schema file which extends the Zowe Component base schema.
 
--- a/docs/appendix/server-component-manifest.md
+++ b/docs/appendix/server-component-manifest.md
@@ -27,6 +27,14 @@ The manifest file contains the following properties:
 - **`license`**
 
   (Optional but recommended) Defines the license code of the component. For example, Zowe core components have `EPL-2.0` value in this field.
+  
+- **`schemas`**
+
+  Defines the location of json schema files that are compatible with certain portions of Zowe as denoted by each child property.
+  
+  * **`config`**
+  
+    Defines the location of the json schema file which extends the Zowe Component base schema.
 
 - **`build`**
 
--- a/docs/extend/packaging-zos-extensions.md
+++ b/docs/extend/packaging-zos-extensions.md
@@ -35,7 +35,7 @@ A typical component package, for example, `jobs-api-package-1.0.4.zip`, consists
    
 - `schema.json`
 
-   Refers to the [json schema](https://json-schema.org/) file that details the parameters that are valid for the component's configuration within Zowe server configuration files. See documentation on [server component schema files](server-schemas.md) for more information.
+   An example filename of the [json schema](https://json-schema.org/) file specified by the manifest property `schemas.configs` as detailed in [Server Component Manifest File Reference](../appendix/server-component-manifest.md). The file details the parameters that are valid for the component's configuration within Zowe server configuration files. See documentation on [server component schema files](server-schemas.md) for more information.
 
 - `apiml-static-registration.yaml.template`
 
--- a/docs/extend/packaging-zos-extensions.md
+++ b/docs/extend/packaging-zos-extensions.md
@@ -20,6 +20,7 @@ A typical component package, for example, `jobs-api-package-1.0.4.zip`, consists
 ```
 +-- manifest.yaml
 |-- apiml-static-registration.yaml.template
+|-- schema.json
 |-- bin/
     |-- configure.sh
     |-- jobs-api-server-1.0.4-boot.jar
@@ -31,6 +32,10 @@ A typical component package, for example, `jobs-api-package-1.0.4.zip`, consists
 - `manifest.yaml`
 
    Refers to the Zowe component manifest file. You can find detailed definition of manifest in [Server Component Manifest File Reference](../appendix/server-component-manifest.md).
+   
+- `schema.json`
+
+   Refers to the [json schema](https://json-schema.org/) file that details the parameters that are valid for the component's configuration within Zowe server configuration files. See documentation on [server component schema files](server-schemas.md) for more information.
 
 - `apiml-static-registration.yaml.template`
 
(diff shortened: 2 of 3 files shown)


samanthasusu (approved):

docs/extend/packaging-zos-extensions.md
```diff
@@ -31,6 +32,10 @@ A typical component package, for example, `jobs-api-package-1.0.4.zip`, consists
 - `manifest.yaml`
 
    Refers to the Zowe component manifest file. You can find detailed definition of manifest in [Server Component Manifest File Reference](../appendix/server-component-manifest.md).
+   
+- `schema.json`
+
+   Refers to the [json schema](https://json-schema.org/) file that details the parameters that are valid for the component's configuration within Zowe server configuration files. See documentation on [server component schema files](server-schemas.md) for more information.
```
jackjia-ibm, line 38: Is it worthy to clarify that`schema.json` file name is customizable by defining in `manifest.yaml` `schemas.configs` section.

docs/appendix/server-component-manifest.md
```diff
@@ -27,6 +27,14 @@ The manifest file contains the following properties:
 - **`license`**
 
   (Optional but recommended) Defines the license code of the component. For example, Zowe core components have `EPL-2.0` value in this field.
+  
+- **`schemas`**
```
jackjia-ibm, line 31: I'm trying to think of the case if this should be plural or singular. Any examples we can think of other schemas could come in? But plural could be the safe case.

docs/appendix/server-component-manifest.md
```diff
@@ -27,6 +27,14 @@ The manifest file contains the following properties:
...
+
+  Defines the location of json schema files that are compatible with certain portions of Zowe as denoted by each child property.
+  
+  * **`config`**
```
jackjia-ibm, line 35: Is that ok to make it plural so it aligns the `configs` section in manifest.yaml. My understanding this is the schema to describe `configs` sections. Not sure if I understand correctly.

docs/extend/server-schemas.md
```diff
@@ -0,0 +1,64 @@
...
+
+The server infrastructure will utilize each components' schema files to validate a Zowe instance configuration every startup, so this requirement is enforced by code.
+
+## Requirements
+
+* Server component json schema files must follow the json schema [spec 2019-09](https://json-schema.org/draft/2019-09/schema).
+* Each component must state where its base schema file is located by the manifest parameter "schemas.config"
+* The schema file must use and/or extend the Zowe Component base schema by use of the "allOf" attribute.
+* The schema must have an `$id` property which is a URI that has a domain related to the entity that developed the Component.
+* The file should be tagged on z/OS but elsewhere must at least be encoded as ASCII-subset of UTF-8
```
jackjia-ibm, line 13: We don't have tagging requirement on any other files. Any particular reason this schema file must be tagged?

docs/extend/server-schemas.md
```diff
@@ -0,0 +1,64 @@
...
+### Example manifest
+
+```
+name: component1
+id: com.abcdef.component1
+title: component1
+description: An example component with a simple configuration.
+license: ABCDEF Company License
+schemas:
+  - config: schema.json
```
jackjia-ibm, line 33: Is the `-` needed? Just try to understand the correct format. With dash, it means:

```json
{
  "schemas": [
    {
      "config": "schema.json"
    }
  ]
}
```

docs/extend/server-schemas.md
```diff
@@ -0,0 +1,64 @@
...
+
+Below is an example of the "schema.json" file referenced above. In it, we have 1 special property, "my-custom-prop", which is just a boolean that can be true or false.
+
+```
+{
+  "$schema": "https://json-schema.org/draft/2019-09",
+  "$id": "https://abcdef.com/schema/component1",
+  "allOf": [
+    { "$ref": "https://zowe.org/schema/base" },
+    { "$ref": "https://zowe.org/schema/base#zoweComponent"},
```
jackjia-ibm, line 47: Should we call them `https://zowe.org/schemas/server-component-base` to distinguish with CLI config schema. And with plural `schemas`.

jackjia-ibm: How versioning could be affect this url? I'm not sure what's the best practice here.

jackjia-ibm (commented):
Thanks Sean, I think i understand better how it works now.

docs/appendix/server-component-manifest.md
1000TurquoisePogs: I don't know yet, but I'm making it an object for future-proofing.

docs/appendix/server-component-manifest.md
1000TurquoisePogs: ok makes sense.

docs/extend/packaging-zos-extensions.md
1000TurquoisePogs: yes

docs/extend/server-schemas.md
```diff
@@ -0,0 +1,64 @@
+# Server component schemas
+
+Starting with Zowe v2.0, each Component in Zowe must contain a [json schema](https://json-schema.org/) describing the configuration parameters that are valid for its component section in Zowe's server configuration. If a component does not have anything that can be configured, this file can just be boilerplate specifying that it fully inherits generic Component parameters and nothing more.
```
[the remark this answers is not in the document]
1000TurquoisePogs: the reason to require it is because making a schema with the boilerplate only takes a minute, but not having any enforcement will end up with people who would have been perfectly fine with making a nice schema not even knowing  it was possible to do.

docs/extend/server-schemas.md
1000TurquoisePogs: doesn't need to be, just should be ascii at least. just the way the configmgr reads files.

docs/extend/server-schemas.md
1000TurquoisePogs: oh, i messed this up. yes i'll remove the dash.

docs/extend/server-schemas.md
1000TurquoisePogs: sounds good on name changed.
not sure about versions, checking on this...
i know our schemas will be versioned by github tag but we may need to incorporate version into the URL as json-schema.org does as well.
and uh, these URLs dont need to be real URLs but perhaps we could publish ours there for convenience.

1000TurquoisePogs: checked and its suggested that semver be in the URL

1000TurquoisePogs: i have a concern about the version number in the schema id
basically we should be able to add stuff every zowe release and we can diff to see
but because component schemas need to ref ours, if we change the version number in the id, it will break.
instead we could be a little nonstandard by making a version property separately.

Issue #2112, opened 2022-04-11
Zowe V2 updates for Zowe Explorer


zFernand0:
## Is your request for enhancement related to a problem? Please describe.
Zowe V2 updates for Zowe Explorer

## Describe the solution you'd like
**Zowe Explorer Profiles**
1. I see the CLI squad uses the term `team profiles` to refer to Zowe v2 LTS profiles in [this article](https://deploy-preview-1961--zowe-docs-master.netlify.app/stable/user-guide/cli-using-using-team-profiles/). Maybe we should stick to the following verbiage:
    - For v1: Zowe v1 LTS profiles, v1 profiles, yaml profiles, single profile per yaml file, ...
    - For v2: Zowe v2 LTS profiles, **team** profiles, v2 profiles, ...

2. The section [`working-with-zowe-explorer-profiles`](https://docs.zowe.org/stable/user-guide/ze-profiles/#working-with-zowe-explorer-profiles) should mention that it is mostly applicable for Zowe v1 LTS profile (singe-profile-per-yaml-file).
From this section, the only thing that applies to both, v1/yaml profiles and **team** profiles, is the **Profile Validation**

3. For v2 we should add a Team Configuration file section before the v1 profile creation documentation (above section)
The contents of this new section will include almost everything in the [Profile Configuration section of the Early Access readme file](https://github.com/zowe/vscode-extension-for-zowe/blob/next/docs/Early%20Access%20-%20Using%20Global%20Profile%20Configuration.md#profile-configuration)

4. The [`enabling-secure-credential-store-with-zowe-explorer`](https://docs.zowe.org/stable/user-guide/ze-profiles/#enabling-secure-credential-store-with-zowe-explorer) topic (and its sub heading - for Zowe CLI Users) should be removed. A PR is in the works to remove all of that stuff 😋 

## Related doc pages
Links above

nannanli:
@JamesBauman Can this be closed now?

JamesBauman:
hi @nannanli  ...   I'll get back to you.... The affected article needs a grammar review/update.  Perhaps i should open a "doc enhancement" issue.  Stay tuned... 

thx

PR #2115, opened 2022-04-12
per issue #2112

JamesBauman:
Hi @zFernand0 - 

Please review this PR based on your comments re #2112 

Please base your review on technical accuracy and proper references to global configuration and team config,  After we get the content technically correct, i'll do a deep grammar review.  So please dont approve the PR (yet).

You can preview the file here: https://github.com/zowe/docs-site/blob/jabauman-issue-2112/docs/user-guide/ze-profiles.md

### DIFF
--- a/docs/user-guide/ze-profiles.md
+++ b/docs/user-guide/ze-profiles.md
@@ -2,9 +2,118 @@
 
 After you install Zowe Explorer, you need to have a Zowe Explorer profile to use all functions of the extension. You can optionally activate the Secure Credential Store plug-in to securely store your credentials.
 
-**Note:** You can continue using Zowe V1 profiles with the Zowe Explorer V2. For more information, see ***insert link here***.
+**Note:** You can continue using Zowe V1 profiles with Zowe Explorer V2. 
+## Configuring team profiles
+
+Zowe CLI team profiles simplify profile management by letting you to edit, store, and share mainframe configuration details in one location. You can use a text editor or an IDE to populate configuration files with connection details for your mainframe services. By default, your team configuration file is located in the `.zowe home` folder, whereas the project-level configuration file is located in the main directory of your project.
+
+**Note**: A project context takes precedence over global configuration.
+### Creating team configuration files
+
+Create a team configuration file.
+
+1. Navigate to the explorer tree.
+2. Hover over **DATA SETS**, **USS**, or **JOBS**.
+3. Click the **+** icon.
+4. Select **Create a New Team Configuration File**.
+5. Chose either a global configuration file or a project-level configuration file.
+6. Edit the config file to include the host information.
+7. Refresh Zowe Explorer by either clicking the button in the notification message shown after creation, `alt+z`, or the `Zowe Explorer: Refresh Zowe Explorer` command palette option.
+
+Your team configuration file appears either in your `.zowe` folder if you choose the global configuration file option, or in your workspace directory if you choose the project-level configuration file option. The notification message that shows in VS Code after config file creation will include the path of the file created.
+
+### Managing profiles
+
+You can edit your project-level or global configuration files.
+
+**Follow these steps**:
+
+1. Right-click on your profile.
+2. Select the **Add**, **Update**, or **Delete Profile** options to edit the zowe config file in place.
+
+   **Tip:** Use the Intellisense prompts if you need assistance with filling parameters in the .json file.
+
+3. Refresh the view by clicking the refresh icon in the Data Sets, USS, or Jobs view.
+
+   Alternatively, press F1 to open the command palette, type and execute the **Zowe Explorer: Refresh Zowe Explorer** option.
+
+You successfully edited your configuration file.
+
+### Sample profile configuration
+
+View the profile configuration sample. In the sample, the default `lpar1.zosmf` profile will be loaded upon activation.
+
+You can use the sample to customize your profile configuration file. Ensure that you edit the `host` and `port` values before you work in your environment.
+
+```json
+{
+  "$schema": "./zowe.schema.json",
+  "profiles": {
+    "lpar1": {
+      "properties": {
+        "host": "192.86.32.67"
+      },
+      "profiles": {
+        "zosmf": {
+          "type": "zosmf",
+          "properties": {
+            "port": 10443
+          },
+          "secure": []
+        },
+        "tso": {
+          "type": "tso",
+          "properties": {
+            "account": "",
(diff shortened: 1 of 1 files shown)


docs/user-guide/ze-profiles.md
```diff
@@ -2,10 +2,119 @@
 
 After you install Zowe Explorer, you need to have a Zowe Explorer profile to use all functions of the extension. You can optionally activate the Secure Credential Store plug-in to securely store your credentials.
```
zFernand0, line 3: We should remove the `You can optionally activate the Secure Credential Store plug-in to securely store your credentials.` sentence since it's enabled by default on V2 😋

docs/user-guide/ze-profiles.md
```diff
@@ -2,10 +2,119 @@
...
 
-**Note:** You can continue using Zowe V1 profiles with the Zowe Explorer V2. For more information, see ***insert link here***.
+**Note:** You can continue using Zowe V1 profiles with Zowe Explorer V2. 
+## Configuring team profiles
+
+Zowe CLI team profiles simplify profile management by letting you to edit, store, and share mainframe configuration details in one location. You can use a text editor or an IDE to populate configuration files with connection details for your mainframe services. By default, your team configuration file is located in the `.zowe home` folder, whereas the project-level configuration file is located in the main directory of your project.
+
+**Note**: A project context takes precedence over global configuration.
```
zFernand0, line 10: Should we mention that it's possible to use the global configuration as long as the profile names are different ?

docs/user-guide/ze-profiles.md
```diff
@@ -2,10 +2,119 @@
...
+### Creating team configuration files
+
+Create a team configuration file.
+
+1. Navigate to the explorer tree.
+2. Hover over **DATA SETS**, **USS**, or **JOBS**.
+3. Click the **+** icon.
+4. Select **Create a New Team Configuration File**.
+5. Chose either a global configuration file or a project-level configuration file.
+6. Edit the config file to include the host information.
```
zFernand0, line 20: Kind of obvious, but should we also mention that they need to save the config file before refreshing (next step)? 😋

docs/user-guide/ze-profiles.md
```diff
@@ -2,10 +2,119 @@
...
+### Managing profiles
+
+You can edit your project-level or global configuration files.
+
+**Follow these steps**:
+
+1. Right-click on your profile.
+2. Select the **Add**, **Update**, or **Delete Profile** options to edit the zowe config file in place.
+
+   **Tip:** Use the Intellisense prompts if you need assistance with filling parameters in the .json file.
```
zFernand0, line 34: We should probably avoid saying `.json` file.
Maybe we just say "the file" for now ?
I know this is something we have to review across the Client components for Zowe V2 (after the GA date)

docs/user-guide/ze-profiles.md
```diff
@@ -2,10 +2,119 @@
...
+
+3. Refresh the view by clicking the refresh icon in the Data Sets, USS, or Jobs view.
```
zFernand0, line 36: Again, I assume the word `edit` in the previous step implies that the user should also save the file. 😋 
So, not sure if we need to spell it out 😋

docs/user-guide/ze-profiles.md
```diff
@@ -42,94 +153,37 @@ Follow these steps:
...
 
 4. Select the profile you use with your base profile with the token.
 
    The profile appears in the tree and you can now use this profile to access z/OSMF via the API Mediation Layer.
 
-For more information, see [Integrating with API Mediation Layer](https://docs.zowe.org/stable/user-guide/cli-usingcli#integrating-with-api-mediation-layer).
+For more information, see [Integrating with API Mediation Layer](../user-guide/cli-using-integrating-apiml.md).
 
-### Log in to the Authentication Service
+### Logging in to the Authentication Service
```
zFernand0, line 186: we should be careful using `.yaml` references for content that applies to both v1 and v2 profiles.
`The token is stored in the default base profile .yaml file.` should probably just say `The token is stored in the corresponding base profile.`

docs/user-guide/ze-profiles.md
```diff
@@ -42,94 +153,37 @@ Follow these steps:
...
-4. Add the **Zowe-Plugin** value to the `Zowe Security: Credential Key` entry field.
-5. Restart VSCode.
-6. Create a profile.
-
-Your Zowe Explorer credentials are now stored securely.
-
-### For Zowe CLI users
-
-Ensure that you install the SCS plug-in for Zowe CLI before activating SCS in Zowe Explorer. For more information about the SCS plug-in for Zowe CLI, see [Secure Credential Store plug-in for Zowe Explorer](cli-scsplugin.md).
+1. Open Zowe CLI and run the following command:
```
zFernand0, line 168: Are we trying to illustrate that people can use tokens between CLI and ZE ?
or do just that ZE works with tokens ?

If we only care about SSO through APIML (ZE using tokens) we should probably remove the CLI references from here
(in the future)

zFernand0 (approved):
LGTM! 😋 
I don't think we need to make all of these suggestions though.

docs/user-guide/ze-profiles.md
JamesBauman: done

docs/user-guide/ze-profiles.md
JamesBauman: added "... and save the file." to the end of Step 6.

docs/user-guide/ze-profiles.md
JamesBauman: removed  .json

docs/user-guide/ze-profiles.md
JamesBauman: added new step 3.  Save the config file.

docs/user-guide/ze-profiles.md
JamesBauman: ummm - i see only one ref to CLI in the topic.  Open CLI..... i'm calling this done.

docs/user-guide/ze-profiles.md
JillieBeanSim: I still see this in Files Changed `You can optionally activate the Secure Credential Store plug-in to securely store your credentials.` I think it should be removed since no more SCS plugin

zFernand0:
I'll resolve the comments once the changes are pushed to the branch /PR 😋

PR #2153, opened 2022-04-22
Clean and update release log for API ML

balhar-jakub:
Signed-off-by: Jakub Balhar <jakub.balhar@broadcom.net>

There were some changes to the API ML release log that I believe improve the understandability and remove pieces irrelevant to users and extenders.


janan07 (approved):
Looks good

PR #2282, opened 2022-07-18
created framework for the change log (rel notes)

JamesBauman:
Created the sections for each component in the Release Notes article.

@nannanli - can you please approve?

I created the framework so the we all use the same structure.


nannanli (approved):
Thanks you Jim!

PR #2285, opened 2022-07-19
update to v2.2 webhelp.  includes the extracted files, the PDF, and the zip file

JamesBauman:
@t1m0thyj 

Please review


t1m0thyj (commented):
Was the web help PDF updated from Zowe 2.1 to 2.2? I don't see that it changed in the PR diff.

JamesBauman:
@t1m0thyj  - 

Please recheck and verify...

awharn (approved):

PR #2287, opened 2022-07-19
State java 8 as only supported version for servers

1000TurquoisePogs:
Zowe v2 documentation on server software requirements states that java is required, but it must be version 8.
v1 doc did not say this, it said 8 or above.
The code seems to not work correctly with java 11, so we should make the v1 doc match the v2 doc in requiring specifically java 8 only.


1000TurquoisePogs:
On the TSC call it seemed people were ok with this temporary change, so I'll merge it.

PR #2327, opened 2022-07-29
Fixing dead link

1000TurquoisePogs:
The doc recommends you to do something that we don't recommend and haven't since 2021.
It also points to a non-existing section. If other non-existing sections could be identified I could help to fix more issues like this.


nannanli:
@1000TurquoisePogs Hi Sean, sounds this also applies to previous versions. Could you also remove this prerequisite in `/docs-site/versioned-docs` directory? Thank you!

1000TurquoisePogs:
@nannanli ok I have now done it for every version that is affected.

nannanli (approved):
LGTM

PR #2328, opened 2022-07-29
Fixing broken table

1000TurquoisePogs:
I use the permission table a lot and noticed it was unreadable on docs.zowe.org. 1 line fix to fix it


nannanli (approved):
Looks great! Thank you for making the change.

PR #2349, opened 2022-08-16
Move the API ML related pages to the correct locations. 

balhar-jakub:

The pages for the API ML were mostly under the extend part even for the configurations and user focused topics. This is effort to fix this. 

:heart:Thank you!


janan07 (approved):
Reviewed by Tech Info

PR #2399, opened 2022-09-16
[Zowe 2.3] Adding doc about configmgr

1000TurquoisePogs:
Here is documentation about the `configmgr` that is found in Zowe 2.3 and how you can use it either by itself or via `zwe`. Since this relates to `zwe` I have updated sections about `zwe` to explain `zwe` as early as possible and to link over to `configmgr` documentation as early as possible.


nannanli (approved):
I made some editorial changes. Otherwise looks goff to me. Thank you!

PR #2400, opened 2022-09-19
Add detail to ZWEAG713E

janan07:
This PR is to address issue [#2566](https://github.com/zowe/api-layer/issues/2566) in api-layer


docs/troubleshoot/troubleshoot-apiml-error-codes.md
```diff
@@ -1141,6 +1141,25 @@ The following error message codes may appear on logs or API responses. Use the f
 
   Based on the specific information in the message, verify that the key configuration is correct, or alternatively, that z/OSMF is available.
 
+  Use the following configuration format:
```
balhar-jakub, line 1144: I would add before this line something akin to:

Another potential source of the problem is that API ML doesn't correctly recognize what type of authentication the zOSMF use. In that case it's possible to provide the information for API ML via configuration.

janan07: Ok. I've added a bit of langauge to Reasons and Actions. Thanks for reviewing.

achmelo (approved):

PR #2471, opened 2022-11-02
Git Bash Known Issue doc

anaxceron:
Signed-off-by: anaxceron <ana.ceron@broadcom.com>

- Added the git bash double slash solution to [Known CLI Issues](https://docs.zowe.org/stable/troubleshoot/cli/known-cli)
- Rearranged list of known issues to follow (mostly) alpha order

### DIFF
--- a/docs/troubleshoot/cli/known-cli.md
+++ b/docs/troubleshoot/cli/known-cli.md
@@ -89,7 +89,7 @@ You attempt to issue node.js commands and you do not receive the expected  outpu
 
 There might be a program that is named *node* on your path. The Node.js installer automatically adds a program that is named *node* to your path. When there are pre-existing programs that are named *node* on your computer, the program that appears first in the path is used. To correct this behavior, change the order of the programs in the path so that Node.js appears first.
 
-## `npm install -g `command fails due to an EPERM error
+## `npm install -g` command fails due to an EPERM error
 
 **Valid on Windows**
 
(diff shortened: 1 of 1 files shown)


docs/troubleshoot/cli/known-cli.md
```diff
@@ -116,28 +129,50 @@ issue the following command:
...
-## Node.js commands do not respond as expected
+## Paths converting in Git Bash
 
-**Valid on Windows or Linux**
+**Valid on Windows**
 
 **Symptom:**
 
-You attempt to issue node.js commands and you do not receive the expected  output.
+When issuing a command with a relative directory path using forward slashes, Git Bash converts the path into an absolute path with a root directory.
```
t1m0thyj, line 138: The path that gets converted is an absolute path. We could say something like this:
> When issuing a command with an absolute directory path that begins with a forward slash, Git Bash converts the path into an invalid one that does not exist.

docs/troubleshoot/cli/known-cli.md
```diff
@@ -116,28 +129,50 @@ issue the following command:
...
+  Git Bash converts the path entered in the following command:
 
-**Valid on Oracle Linux 6**
+  ```
+  zowe zos-files upload dir-to-uss local_dir /a/ibmuser/my_dir
+  ```
+  To avoid conversion, the path should be edited as in the following example:
+  ```
+  zowe zos-files upload dir-to-uss local_dir //a/ibmuser/my_dir
+  ```
```
t1m0thyj, line 152: Maybe we should add an example of how the path gets converted and what type of error the user may see.

In this case, the path gets converted to `A:/ibmuser/my_dir` and the user may see a "No such file or directory" error.

docs/troubleshoot/cli/known-cli.md
```diff
@@ -116,28 +129,50 @@ issue the following command:
...
+
+- Use an environment variable in one of the following ways:
+    - In your active terminal, set the environment variable `MSYS_NO_PATHCONV` to `1`.
+    
+      Then add an export of the environment variable once during a terminal session. If writing a script, run the export command once at the top of the script.
```
t1m0thyj, line 157: These two statements are redundant:
* In your active terminal, set the environment variable `MSYS_NO_PATHCONV` to `1`
* Then add an export of the environment variable once during a terminal session

t1m0thyj (commented):
Thanks for documenting this issue 🙂 Left a few comments.

docs/troubleshoot/cli/known-cli.md
anaxceron: per our discussion, will wait on you to provide an example

docs/troubleshoot/cli/known-cli.md
anaxceron: editing to leave in the second sentence

anaxceron:
@t1m0thyj I pushed a new version of the draft content per your feedback. please review when you get the chance.

t1m0thyj (approved):
LGTM 😁

anaxceron:
This content documented the solution for the following issue: #https://github.com/zowe/zowe-cli/issues/1440

PR #2473, opened 2022-11-02
Reorganizing zowe server support pages

1000TurquoisePogs:
This PR makes a category out of the mediation layer, launcher, and appfw support pages to group them together and make it easier to understand that they are related when you have server issues and dont know anything more.

This also creates a new page about information that you should gather when contacting support. its a list of things the cli, webui, and apiml squads already ask for in their issue trackers.

I also resurrected some appfw support pages after cleaning them up to remove v1 content and replace it with v2 content.


Joe-Winchester (approved):
This is great !  I've been on a number of customer calls recently where I have had to duck back to v1 troubleshooting and try to transpose ZWESV to ZWESL and similar, so many thanks for writing the PR.  I like the way the servers are consolidated as well to make Zowe docs align more to the components rather than the squads' contributions.

nannanli (approved):
LGTM ~ thank you for the updates!

Issue #2475, opened 2021-08-18
Cannot chain Zowe CLI commands in a .bat / .cmd script


FALLAI-Denis:
On Windows, either the `test.bat` file with the content:

```
echo 1
zowe --version
echo 2
zowe --version
echo 3
zowe --version
```

The execution of the `test.bat` file stops after the first zowe command...

Checked on different installations with Zowe CLI versions 6.32.0 and 6.32.2 ...

anaxceron:
Closing issue now that we have documentation for this known issue, per [PR 2500](https://github.com/zowe/docs-site/pull/2500).

PR #2500, opened 2022-11-15
added chain commands failure troubleshooting

anaxceron:
Added troubleshooting help content in [Known CLI Issues](https://docs.zowe.org/stable/troubleshoot/cli/known-cli) for [Issue 2475](https://github.com/zowe/docs-site/issues/2475)

### DIFF
--- a/docs/troubleshoot/cli/known-cli.md
+++ b/docs/troubleshoot/cli/known-cli.md
@@ -43,7 +43,7 @@ This behavior occurs under the following conditions:
 
 **Solution:**
 
-To prevent the Zowe executable from taking control and stopping the execution of the script it is called from, add `call` in front of the first Zowe CLI command.
+To prevent the Zowe executable from taking control and stopping the execution of the script it is called from, add `call` in front of each Zowe CLI command.
 
 In the example below, the script never prints "hello" to the terminal:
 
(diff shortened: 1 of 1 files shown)


docs/troubleshoot/cli/known-cli.md
```diff
@@ -35,6 +35,29 @@ This behavior occurs under the following conditions:
...
 
+## Chain commands fail in a batch script
+
+**Valid on Windows**
+
+**Symptom:** When invoking Zowe CLI in a batch script (`.bat` or `.cmd` file), subsequent commands in the script do not run.
+
+**Solution:**
+
+To prevent the Zowe executable from taking control and stopping the execution of the script it is called from, add `call` in front of the first Zowe CLI command.
```
t1m0thyj, line 46: Instead of "the first Zowe CLI command", would it make sense to say "each Zowe CLI command"?

Because if there are multiple Zowe CLI scripts, I'd need to add `call` before each one. For example:
```batch
call zowe --version
echo hello
call zowe --version
echo goodbye
```

anaxceron: ah, yes, this needs to be clarified. I'll update now. (btw, not sure why the preview build is failing)

anaxceron: @t1m0thyj  - updated. please review when you get a chance.

t1m0thyj (approved):

zFernand0 (approved):

PR #2504, opened 2022-11-17
Standardize zowe architecture page

1000TurquoisePogs:
The architecture page has some info that's a little out of date, but also a little misleading versus what we often want to tell our users.
I've not only updated the webui sections, but added sections for each component that standardizes some important knowledge: where to find the servers, and what their jobnames are.

### DIFF
--- a/docs/getting-started/zowe-architecture.md
+++ b/docs/getting-started/zowe-architecture.md
@@ -52,17 +52,30 @@ When deploying other server components into container orchestration software lik
 - Components may register themselves to Discovery with their own `Pod` name within the cluster.
 - Zowe workloads use the `zowe-launch-scripts` `initContainers` step to prepare required runtime directories.
 - Only necessary components ports are exposed outside of Kubernetes with `Service`.
+
 ## App Server
 
-The App Server is a node.js server that is responsible for the Zowe Application Framework. This server provides the Zowe desktop, which is accessible through a web browser via port 7556. The Zowe desktop includes a number of applications that run inside the Application Framework such as a 3270 emulator and a File Editor.
+The App Server is a portable, extensible HTTPS server written in node.js. It can be extended with expressjs routers to add REST or Websocket APIs. This server is responsible for the Zowe Application Framework, including the Desktop which is described later in this page.
 
 ![Zowe Desktop Diagram](../images/mvd/zowe-desktop.png)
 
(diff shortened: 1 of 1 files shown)


docs/getting-started/zowe-architecture.md
```diff
@@ -52,41 +52,67 @@ When deploying other server components into container orchestration software lik
...
 - Only necessary components ports are exposed outside of Kubernetes with `Service`.
+
 ## App Server
 
-The App Server is a node.js server that is responsible for the Zowe Application Framework. This server provides the Zowe desktop, which is accessible through a web browser via port 7556. The Zowe desktop includes a number of applications that run inside the Application Framework such as a 3270 emulator and a File Editor.
+The App Server is a portable, extensible HTTPS server written in node.js. It can be extended with expressjs routers to add REST or Websocket APIs. This server is responsible for the Zowe Application Framework, including the Desktop which is described later in this page.
 
 ![Zowe Desktop Diagram](../images/mvd/zowe-desktop.png)
 
-The App Server server logs write to `<zowe.logDirectory>/appServer-yyyy-mm-dd-hh-mm.log`.  The Application Framework provides REST APIs for its services that are included on the API catalog tile `Zowe Application Framework` that can be viewed at `https://<ZOWE_HOST_IP>:7554/apicatalog/ui/v1/#/tile/ZLUX/zlux`.
```
DivergentEuropeans, line 61: Surely it's okay to keep 
"The App Server server logs write to `<zowe.logDirectory>/appServer-yyyy-mm-dd-hh-mm.log`."
even if it's redundant info somewhere else, for user's convenience sake?

docs/getting-started/zowe-architecture.md
```diff
@@ -52,41 +52,67 @@ When deploying other server components into container orchestration software lik
...
+When the API gateway is running, this server and the Desktop are accessible at `https://<ZOWE_HOST_IP>:7554/zlux/ui/v1/`.
```
DivergentEuropeans, line 62: 7554 is default. Probably best to

When the API gateway is running, this server and the Desktop are accessible at `https://<zowe.externalDomains[0]>:<zowe.externalPort>/zlux/ui/v1/`.

Note: Default Zowe external port (Gateway port) is 7554

DivergentEuropeans (approved):
Approving because this is an improvement 100%, but I think my comments are worth addressing

docs/getting-started/zowe-architecture.md
1000TurquoisePogs: I don't find that useful in a page titled "architecture". it's in relevant places (advanced, troubleshooting);

docs/getting-started/zowe-architecture.md
1000TurquoisePogs: That'd require explaining zowe.yaml on this page, which we don't do. The numbers we see match the diagram.

PR #2510, opened 2022-11-21
Update troubleshooting tip to state that zOSMF user is workaround. 

balhar-jakub:
Signed-off-by: Jakub Balhar <jakub.balhar@broadcom.com>

The troubleshooting tip state incorrectly that the Zowe should be run under the same user as zOSMF which isn't true. It's just possible workaround for the specific problem. 

:heart:Thank you!


janan07 (approved):
Reviewed and edited by Tech Info

PR #2527, opened 2022-12-05
Fix confusion of keyring with keystore

1000TurquoisePogs:
In the page that I've fixed, 'keystore' is written when 'keyring' is intended.
A keyring is a keystore. But there are other types of keystores. The entire page talks about the other type, pkcs12 files. There's a different page on the keyring topic too. So, the occurences of keystore that I changed to keyring I changed because contextually it is understood that the page is contrasting one type of keystore to the other type (keyring).


rpenny125 (approved):
I approve (for what it's worth) the specific change from "keystores" to "keyrings". I leave comment on possible additional issues to others.

PR #2571, opened 2023-01-04
appserver message descriptions

1000TurquoisePogs:
This PR contains a description for every message with an ID that can be printed by the app server.
It will help people to find the reason for a message and what actions they should take.


lparadkar-rocket (approved):

lparadkar-rocket:
@1000TurquoisePogs As discussed, we can go ahead with the changes.

PR #2580, opened 2023-01-10
fix heap64 suggestion

1000TurquoisePogs:
Madeleine Dudek noted that the current heap64 section isnt actually valid compared to the documentation of ibm here https://www.ibm.com/docs/en/zos/2.4.0?topic=options-heap64-amode-64-only
This PR fixes it, it was obvious what was intended so the fix is simple.


1000TurquoisePogs:
I've reduced the heap64 numbers in light of a user who tried to use the numbers we suggested and ran out of space. Apparently the initial heap sizes are aggressive and that was leading to out of memory issue.

Joe-Winchester (approved):

PR #2581, opened 2023-01-10
Update extend-sdks.md

zFernand0:
Updates from:
- Dec 13th
- Jan 10th
- Jan 17th

View latest changes: https://github.com/zowe/docs-site/blob/zFernand0-patch-1/docs/extend/extend-sdks.md


zFernand0:
Please see: https://github.com/zowe/docs-site/blob/draft-sdkcontent-sdks/docs/extend/extend-sdks.md

PR #2587, opened 2023-01-13
using cURL for troubleshooting CLI

anaxceron:
Created a new page to document using cURL for troubleshooting, including the following topics:
- Understanding cURL commands
- Comparing cURL and Zowe CLI commands

### DIFF
--- a/docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
+++ b/docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
@@ -4,6 +4,12 @@ When a REST API call fails on Zowe CLI, you can use cURL to troubleshoot.
 
 Run a command with cURL and compare its output with what is returned using Zowe CLI. This can help pinpoint whether the problem lies with z/OSMF or Zowe CLI, depending on which command returns an API error.
 
+## Installing cURL
+
+cURL is installed by default on Windows 10 V1803 and later, macOS, and most Linux distributions.
+
+If the cURL command is missing from your system, you can download it from the cURL [Releases and Downloads](https://curl.se/download.html) page.
+
 ## Understanding cURL commands
 
 A cURL command using the `https` protocol follows the basic syntax `curl <URL>`.
@@ -20,21 +26,21 @@ curl --location --request <API method> "https://<host>:<port>/<API>" --header "X
 
 Use `--location` to allow the server to redirect to a different URL, if needed. 
 
-When a different location exists and `--location` is not included in the command, the server responds with a 3XX status code.
+When the server attempts to redirect and `--location` is not included in the command, the server responds with a 3XX status code.
 
 ### `--request <API method>`
 
-Use `--request` to identify the API method and name. Not necessary when the API method is `GET`.
+Use `--request` to identify the API method (such as `POST`, `GET`, `PUT`, `DELETE`). Not necessary when the API method is `GET`.
 
 - `<API method>`: Specifies the API method used in the request.
 
-### `https://<host>:<port>/<API>"`
+### `"https://<host>:<port>/<API>"`
 
 Indicates the protocol and URL.
 
 - `<host>`: Specifies the host name where the z/OSMF services are running.
-- `<port>`: Specifies the REST port number.
-- `<API>`: Specifies the name of the API used in the request.
+- `<port>`: Specifies the REST port number. If not specified, defaults to 443 for HTTPS.
+- `<API>`: Specifies the API endpoint used in the request.
 
 ### `--header "X-CSRF-ZOSMF-HEADER;"`
 
@@ -44,7 +50,7 @@ Required to establish communication with z/OSMF. Specifies that the client is se
 
 ### `--insecure`
 
-Use `--insecure` with a trusted server that does not require verification before a data transfer.
+Use `--insecure` to bypass SSL certificate validation if you are connecting to a trusted server that has self-signed certificates.
 
 ### `--user "<ID>:<PASSWORD>"`
 
@@ -52,6 +58,8 @@ Required and displays as plain text. Also possible to [use an environment variab
 
 - `<ID>`: Specifies the z/OSMF user identification.
 - `<PASSWORD>`: Specifies the user password for z/OSMF.
+
+**NOTE:** To be prompted for a password instead of displaying it in plain text, omit the password from the command and enter `--user "<ID>"`.
 
 ## Comparing commands
 
--- a/docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
+++ b/docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
@@ -96,6 +96,6 @@ z/OSMF Plug-ins that are installed on 'lpar.hostname.net':
 msgId:   IZUG612E
 msgText: IZUG612E
 ```
-When both responses include a matching error, that may indicate there could be a problem with z/OSMF.
+When both responses include a the same error, that may indicate there could be a problem with z/OSMF.
 
 If an API call fails with the Zowe CLI command but not cURL, this can mean the problem lies with Zowe CLI.
--- a/docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
+++ b/docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
@@ -1,0 +1,101 @@
+# Using cURL to troubleshoot Zowe CLI
+
+When a REST API call fails on Zowe CLI, you can use cURL to troubleshoot.
+
+Run a command with cURL and compare its output with what is returned using Zowe CLI. This can help pinpoint whether the problem lies with z/OSMF or Zowe CLI, depending on which command returns an API error.
+
+## Understanding cURL commands
+
+A cURL command using the `https` protocol follows the basic syntax `curl <URL>`.
+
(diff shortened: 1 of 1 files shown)


docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
```diff
@@ -0,0 +1,101 @@
...
+
+### `--location`
+
+Use `--location` to allow the server to redirect to a different URL, if needed. 
+
+When a different location exists and `--location` is not included in the command, the server responds with a 3XX status code.
+
+### `--request <API method>`
+
+Use `--request` to identify the API method and name. Not necessary when the API method is `GET`.
```
t1m0thyj, line 27: "and name" seems redundant here.

docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
```diff
@@ -0,0 +1,101 @@
...
+
+- `<API method>`: Specifies the API method used in the request.
```
t1m0thyj, line 29: We could add examples for the most commonly used ones:
> Specifies the API method used in the request. For example: `POST`, `GET`, `PUT`, `DELETE`

docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
```diff
@@ -0,0 +1,101 @@
...
+
+### `https://<host>:<port>/<API>"`
```
t1m0thyj, line 31: Should the trailing quote be removed?

docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
```diff
@@ -0,0 +1,101 @@
...
+
+Indicates the protocol and URL.
+
+- `<host>`: Specifies the host name where the z/OSMF services are running.
+- `<port>`: Specifies the REST port number.
+- `<API>`: Specifies the name of the API used in the request.
```
t1m0thyj, line 37: Instead of "name" should we say "endpoint"?

docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
```diff
@@ -0,0 +1,101 @@
...
+curl --location --request <API method> "https://<host>:<port>/<API>" --header "X-CSRF-ZOSMF-HEADER;" --insecure --user "<ID>:<PASSWORD>"
+```
+**NOTE:** Some terminals may require single quotes rather than double quotes.
+
+
+### `--location`
+
+Use `--location` to allow the server to redirect to a different URL, if needed. 
+
+When a different location exists and `--location` is not included in the command, the server responds with a 3XX status code.
```
t1m0thyj, line 23: Instead of "When a different location exists" would it be clearer to say "When the server attempts to redirect"?

docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
```diff
@@ -0,0 +1,101 @@
...
+
+- `<host>`: Specifies the host name where the z/OSMF services are running.
+- `<port>`: Specifies the REST port number.
+- `<API>`: Specifies the name of the API used in the request.
+
+### `--header "X-CSRF-ZOSMF-HEADER;"`
+
+Required to establish communication with z/OSMF. Specifies that the client is sending a cross-site request to the REST interface.
+
+- `;`: Indicates that the header has no value. (A value is not required.)
```
t1m0thyj, line 43: We could add a note about how to pass headers with values like this:
> If you need to pass additional headers with values, use a colon to separate the key and value. For example: `--header "X-IBM-Data-Type: binary"`

docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
```diff
@@ -0,0 +1,101 @@
...
+
+### `--insecure`
+
+Use `--insecure` with a trusted server that does not require verification before a data transfer.
```
t1m0thyj, line 47: I would suggest to mention SSL certificates here, maybe something like the following:
> Use `--insecure` to bypass SSL certificate validation if you are connecting to a trusted server that has self-signed certificates.

docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
```diff
@@ -0,0 +1,101 @@
...
+
+### `--user "<ID>:<PASSWORD>"`
+
+Required and displays as plain text. Also possible to [use an environment variable](../../user-guide/cli-using-using-environment-variables.md).
+
+- `<ID>`: Specifies the z/OSMF user identification.
+- `<PASSWORD>`: Specifies the user password for z/OSMF.
```
t1m0thyj, line 54: We could add a note here:
> To be prompted for your password instead of displaying it in plain text, omit the password from the command and specify `--user "<ID>"`.

docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
```diff
@@ -0,0 +1,101 @@
+# Using cURL to troubleshoot Zowe CLI
+
+When a REST API call fails on Zowe CLI, you can use cURL to troubleshoot.
```
t1m0thyj, line 3: Users may be unsure whether they have cURL installed. Should we add something like the following to the introductory paragraph?
> cURL is installed by default on Windows 10 versions 1803 and later, macOS, and most Linux distros. If the `curl` command is missing from your system, you can download it [here](https://curl.se/download.html).

docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
```diff
@@ -0,0 +1,101 @@
...
+Use `--request` to identify the API method and name. Not necessary when the API method is `GET`.
+
+- `<API method>`: Specifies the API method used in the request.
+
+### `https://<host>:<port>/<API>"`
+
+Indicates the protocol and URL.
+
+- `<host>`: Specifies the host name where the z/OSMF services are running.
+- `<port>`: Specifies the REST port number.
```
t1m0thyj, line 36: We could add a note about the default value for port:
> Specifies the REST port number. If not specified, defaults to 443 for HTTPS.

t1m0thyj (commented):
This article will be a very useful resource and I like the way it is organized, very comprehensive and easy to follow 🙂 Left a few comments

docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
anaxceron: Are values actually needed for these cURL commands? If not, then I don't think we need to explain how to include values.

t1m0thyj: No values are needed for the cURL command to check z/OSMF status in the example below. But header values may be required for other cURL commands. For example to upload a binary file using the z/OSMF API you would need to provide the `"X-IBM-Data-Type: binary"` header.

docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
```diff
@@ -0,0 +1,109 @@
...
+```
+{"zos_version":"04.28.00","zosmf_port":"443","zosmf_version":"28","zosmf_hostname":"lpar.hostname.net","plugins":{"msgId":"IZUG612E","msgText":"IZUG612E"},"zosmf_saf_realm":"SAFRealm","zosmf_full_version":"28.0","api_version":"1"}
+```
+
+**Submitting the Zowe CLI command:**
+
+Run the following example command using your information:
+
+```
+zowe zosmf check status --host lpar.hostname.net --port 443 --user ibmuser --password password
```
t1m0thyj, line 90: To make the CLI behave identically to the `-insecure` flag of the cURL command, we should add the `--reject-unauthorized false` option.

docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
zFernand0: This is a bit more than just bypassing self-signed certificates though.
I know Dan has always mentioned this 😋 
maybe if we could use the self-signed certs as an example of the things it won't do 😅

docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
zFernand0: Instead of using version numbers, should we try to use the date format for the win10 update?
That way we can change the wording to "Windows 10  after the April 2018 update" ?

t1m0thyj: I think we should pick whichever is easier for users to check. Personally I like the version number because I can go to System Settings and see whether I have that version number or not. I don't remember what date I last updated Windows. But maybe I'm weird in this regard 😀

zFernand0 (approved):
Overall: LGTM! 😋

docs/troubleshoot/cli/cli-use-curl-to-troubleshoot.md
anaxceron: I think using the v number is better in this case

t1m0thyj (approved):

PR #2648, opened 2023-02-13
server return codes explanation

1000TurquoisePogs:
This adds 2 new documents.
1 page which explains how to find the meaning of generic return codes NOT originating from Zowe servers, but instead from system utilities or Zowe startup programs.
Then it adds another page on app-server specific return codes.

I also found the verify-fingerprint page to be duplicated on the troubleshoot section so fixed that too.


janan07 (approved):
Reviewed by Tech Info

PR #2660, opened 2023-02-17
Fix infinite loop in "docs:pdf" script

t1m0thyj:
Thanks to @awharn for helping to debug and fix these issues 🙂 

* Fixes infinite loop in "docs:pdf" script caused by `sidebars.js` referencing the same article twice
* ~Restores file `api-mediation-security.md` that seems to have been accidentally removed causing broken links~
  Determined this file was removed intentionally, the broken links will be fixed in a subsequent PR
* ~Updates dependencies to fix some vulnerabilities~
  Reverted since it seemed to be making the Netlify build time out

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -610,7 +610,6 @@ module.exports = {
             "troubleshoot/launcher/launcher-error-codes"
           ],
         },
-        "troubleshoot/verify-fingerprint",
       ],
     },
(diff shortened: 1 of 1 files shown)


anaxceron (approved):
Thanks for fixing this so quickly! Looks great

PR #2667, opened 2023-02-21
Zowe z/OS components installation checklist

janan07:
This PR is to introduce the installation checklist of Zowe z/OS components to the section Installing z/OS Components.


janan07:
One option would be to create two installation checklists: One for development purposes and another for production purposes. 

Another consideration is to list the results of performing the various tasks. 
The columns for 'role' and 'status' could be removed to allow for a single column which would show the results of each task.

balhar-jakub (approved):

anaxceron (commented):
I think the Preparing and Installing  and Configuring Zowe z/OS Components tables have too many links. I'd try to place links only under the Task column, like you have done in the Verify Zowe installation on z/OS table. (Also: standardize your capitalization across table names.)

I think having steps broken out into numbers and letters makes this list look too complicated. I would place some of your Description steps as steps in the Task column. In the Configuring Zowe z/OS Components table, I would make "Initialize the z/OS system" a task and items a-e as items in the Description column. I also think it's important that your steps match the titles to the articles you're linking to. (Meaning, if the step is How to Pet your Cat, the linked article should have the same title.) I also think the descriptions in the Configuring Zowe z/OS Components table are not really describing anything but are, instead, lists of more tasks to do.

I would remove the Status column on the webpage, but would consider leaving it in the Excel spreadsheet.

For the Time Estimate column, I would consider another metric, or increment. Maybe it's more valuable to say a step will take "more time" or "less time" -- or "a few hours", "less than an hour", or "about a day." Obviously, if you do take this route (which is what I'm considering doing for _my_ CLI Installation Checklist), you need to define/explain this somewhere on this page.

My overall thought is that there is too much going on here. Gathering all this information in one place is certainly valuable, but you need to help guide the reader through it. I think we need consistency here (links _only_ in the Task column, or _only_ in the Description column), we need better descriptions, and, yes, I also think we need to separate out checklists for specific use cases.

Good job harnessing all this information, I can tell you've been very thorough. I know that wasn't easy.

jalel01 (commented):
Under 'Plan the Installation' we have z/OSMF HA as required for production. Isn't it more recommended? Otherwise, we are saying that you must have HA configured for production which i think is more of a decision on the customer side.

anaxceron (commented):
Two things: 
1. Check your capitalization. Zowe Docs style follows sentence-style capitalization for headings.
2. I still think you have too many links in your tables. I'd make the links in your Description cells the steps in your Task cells.

pablocarle (approved):
This looks good to me

PR #2586, opened 2023-01-13
added troubleshooting tips to the zos-certificate section

samanthasusu:
Signed-off-by: samanthasusu <Wen.Ting.Su@ibm.com>
- added the troubleshooting tips to the z/os certificate troubleshooting guide

### Your checklist for this pull request
🚨Please review the [guidelines for contributing](../docs/contribute/contributing.md) to this repository.

- [ ] If the changes in this PR is part of the next future release, make this pull request against the **docs-staging** branch which will be published at the next release boundary. If the changes in this PR are part of the current release, use the default base branch, **master**. For more information about branches, see https://github.com/zowe/docs-site/tree/master#understanding-the-doc-branches. 
      
- [ ] If this PR relates to GitHub issues in `docs-site` or other repositories, please list in Description, prefixed with **close**, **fix** or **resolve** keywords.

### Description (including links to related git issues)
Please describe your pull request.
- added the troubleshooting tips to the z/os certificate troubleshooting guide
- resolve #2459 
:heart:Thank you!

PR #2675, opened 2023-02-22
Applying 2459 troubleshooting tip to v2

1000TurquoisePogs:
https://github.com/zowe/docs-site/pull/2586 had applied troubleshooting tips to v1 but the same applies to v2.

I made some additional changes:

1) the troubleshooting page for zos certificates was missing from the sidebar, so i restored it
2) the page had v1 specific info, so i removed that.
3) i changed the suggestion to downgrade to "java 7" to "java 8 sr7 (fp10)". i believe https://github.com/zowe/docs-site/pull/2586 may be in error due to that, because I'm not sure who has java 7 these days?


samanthasusu (approved):
LGTM! Will also merge the changes to the master branch later.

Issue #2515, opened 2022-11-22
Issue with docs.zowe.org/stable/user-guide/cli-using-using-profiles


COBOL-Erik:
## Description
zowe profiles commands are no longer valid, yet the documentation refers to zowe profiles commands and not zowe config commands.

## Pages to Update
docs.zowe.org/stable/user-guide/cli-using-using-profiles

## Additional context
It is unclear how a Zowe CLI user should work with profiles now in v2.

anaxceron:
Closing w/ [PR 2732](https://github.com/zowe/docs-site/pull/2732)

PR #2732, opened 2023-03-14
updating team config documentation

anaxceron:
Updated/corrected team config documentation:

- Renamed **Using profiles** to **Using V1 profiles**
- Renamed section header **Configuring team profiles** to **Using team profiles**
- Renamed **Using team profiles** to **Team configurations**
- Renamed **Understanding profile configurations in Zowe CLI** to **How Zowe CLI uses configurations**
- Edited content as needed
- Reordered ToC as needed

Addresses deprecated commands cited in [Issue 2515](https://github.com/zowe/docs-site/issues/2515)


docs/user-guide/cli-using-using-profiles-v1.md
zFernand0, line 1: There is a problem with navigating between versions after a change in the file path (name or folder location).
Is there something we can do about this?
For instance, if I'm viewing this article and I switch to 2.5.x, it will take me back to the overview page since the `-v1` wasn't part of the URI before.

**Note:** This is not something we need to resolve in this PR.

zFernand0 (approved):
LGTM! 😋 

In general, I do agree with switching the existing `Using profiles` to `Using V1 profiles`.
That way is clearer that you are using old configurations.

Left a small question, but nothing that needs resolved in this PR.

Also, thanks for fixing some of the broken links 🥳

adam-wolfe (approved):

docs/user-guide/cli-using-using-profiles-v1.md
anaxceron: I will make a note about this and address in the future -- thanks for bringing this up.

PR #2756, opened 2023-03-20
Update documentation about logLevels

1000TurquoisePogs:
The existing documentation about app framework log levels is slightly confusing to people, so I've tried to make it more obvious how things link to each other and to avoid pitfalls.


balhar-jakub:
Just as a question, do you know whether we have an easy way to say I want to have as detailed logs for all components of all projects as possible?

anaxceron:
@1000TurquoisePogs -- do we really need a "Note" here, or can we simply use this sentence as the introduction to this section? It's not clear which attributes this sentence refers to.

image

PR #2757, opened 2023-03-21
Add troubleshooting ssl tls setup to apiml troubleshooting

janan07:
This PR is to add a troubleshooting section for SSL/TLS setup and links to the corresponding blog post.


balhar-jakub:
As the first step to get the article to the documentation, this is good.

PR #2764, opened 2023-03-23
generating v2.7.0 Zowe Docs PDF and archiving v2.2

anaxceron:
Creating the new Zowe Docs PDF for v2.7.0


zFernand0 (approved):
LGTM! :yum:

Looks like another exit code 137 (out of memory)

PR #2774, opened 2023-03-27
Anax v270 suppotenvs cli

anaxceron:
Adding list of supported environments, resource managers for Zowe CLI.

Also alphabetized similar list for Zowe Explorer.


docs/user-guide/ze-usage.md
```diff
@@ -6,23 +6,24 @@ Review this section to familiarize yourself with Zowe Explorer and make the best
...
+- Windows 10+
 
-### Integrated development environments:
+### Integrated development environments: 
 
-- [Theia](https://theia-ide.org/) 18+
-- [Red Hat CodeReady Workspaces](https://www.redhat.com/en/technologies/jboss-middleware/codeready-workspaces)
 - [Eclipse Che](https://www.eclipse.org/che/)
+- [Red Hat CodeReady Workspaces](https://www.redhat.com/en/technologies/jboss-middleware/codeready-workspaces)
+- [Theia](https://theia-ide.org/) 18+
```
t1m0thyj, line 21: Theia has not reached version 18 yet, I think this should say `1.18+` to match the ZE readme: https://github.com/zowe/vscode-extension-for-zowe/blob/cea37e20cf41dc4073c63f7c0bb652a019472bfe/packages/zowe-explorer/README.md?plain=1#L31

docs/user-guide/ze-usage.md
```diff
@@ -6,23 +6,24 @@ Review this section to familiarize yourself with Zowe Explorer and make the best
...
 
 ### Platform versions:
```
t1m0thyj, line 23: I would suggest removing this "Platform versions" section unless others object:
* Node.js and Yarn do not apply - they are probably left over from a section of developer requirements. For end users of ZE, they shouldn't have to care about either of these.
* VS Code is another development environment (alternative to Theia or Che) and could be moved to the list above. Perhaps as the first item in that list? I know the lists are mostly alphabetized, but VS Code kind of has priority as the primary environment.

t1m0thyj (changes requested):

docs/user-guide/ze-usage.md
```diff
@@ -6,23 +6,24 @@ Review this section to familiarize yourself with Zowe Explorer and make the best
 
 ### Operating systems
 
-- Windows 10+
+
 - MacOS 10.15 (Catalina), 11 (Big Sur), 12 (Monterey)
 - Unix-like:
    - [CentOS](https://www.centos.org/) 8+
    - [RHEL](https://www.redhat.com/en/technologies/linux-platforms/enterprise-linux) 8+
    - [Ubuntu](https://ubuntu.com/) 20.04+
```
awharn, line 14: We may want to also add that we support Debian 11 (bullseye). And for all of these operating systems, we need to disclaim that we only support x64, and features may not function correctly on other CPU architectures (i.e. x86, arm64, MIPS, s390(x)...)

As far as I am aware, the CLI works in most of those environments, but Daemon mode does not.

docs/user-guide/ze-usage.md
t1m0thyj: Should we also add a note similar to the one we recently added to the ZE readme to recommend community releases?
> Zowe Explorer is compatible only with Theia 1.18.0 or higher. We recommend using a [Theia community release](https://theia-ide.org/releases/) as Zowe Explorer could experience possible unexpected behaviors with the latest Theia releases.

docs/user-guide/cli-using-usingcli.md
```diff
@@ -1,7 +1,35 @@
...
+- Unix-like:
+   - [CentOS](https://www.centos.org/) 8+
+
+   - [RHEL](https://www.redhat.com/en/technologies/linux-platforms/enterprise-linux) 8+
+   - [Ubuntu](https://ubuntu.com/) 20.04+
+
+- Windows 10+
+- [z/OS Unix Systems Services](https://www.ibm.com/docs/en/zos/2.4.0?topic=descriptions-zos-unix-system-services) <sup>*</sup>
+
+    <b>*</b> Zowe CLI does not support secure credential storage in this environment.
```
gejohnston, line 23: Question for the team: If a customer is ok with plain text creds, are we currently in a position to officially support everything else under USS? I did not think that we had reached that point.

gejohnston (changes requested):

docs/user-guide/ze-usage.md
anaxceron: Updated the file to reflect both comments -- thank you!

docs/user-guide/cli-using-usingcli.md
awharn: As far as I am aware, DB2 support should exist if the instructions in ibm_db are followed, which as far as I am aware, is not documented anywhere in the Zowe space. The daemon is also not going to work, and I do not think there is any way that we will be able to get that to work. Aside from those, the node CLI should work without any issues.

docs/user-guide/ze-usage.md
awharn: To complicate matters, we also need to disclaim that we do support running on MacOS with the M1/M2/M3 processor with Rosetta, UNLESS they want the DB2 plug-in, in which case the Intel version of Node is required.

docs/user-guide/cli-using-usingcli.md
gejohnston: My reason for this question is that there is a difference between "most stuff works" and "we support this environment". For example, are we prepared to accept and fix an issue reported on USS (even in the areas that we think should work) with an acceptable service level response time? Is USS part of our automated test suites? Are we confident that we are ready to support USS?

I added this question as a topic for the last Product Management meeting. However, the agenda was packed and we did not get to this question. The question remains on the agenda for the next meeting.

anaxceron: Team consensus: Remove USS from list

To include USS in the future: Document install, complete list of items that don't work on USS (sec cred, daemon, etc.); internal testing functionality

adam-wolfe: On this page (cli-using-usingcli.md), we could suggest that people who are interested in Zowe CLI on USS visit:  https://github.com/zowe/zowe-cli/issues/1680 and upvote (👍) to show interest.

docs/user-guide/ze-usage.md
awharn: Well, this thread was meant to be about the CLI, and I appear to have accidentally put it on the Zowe Explorer and cannot move the conversation. Oops. :/

I think, if we have a section, we are just going to say that the CLI supports x86_64 based processors on Windows 10, Ubuntu 20.04+, Debian 11+, RHEL 8+, CentOS 8+, and MacOS 10.15+.

docs/user-guide/ze-usage.md
anaxceron: Updated the content as Timothy advised - thank you!

docs/user-guide/ze-usage.md
anaxceron: Feedback incorporated into the cli-using-usingcli.md file -- thank you!

docs/user-guide/cli-using-usingcli.md
anaxceron: Feedback incorporated into the content. Thank you!

docs/user-guide/cli-using-usingcli.md
```diff
@@ -1,7 +1,35 @@
...
+### Package/resource managers
+
+- [NodeJS](https://nodejs.org/en)
+- [npm](https://www.npmjs.com/) 6+
+- [PnPM](https://pnpm.io/)
+- [Yarn](https://yarnpkg.com/) 
+
+<br/> 
+
+>Using Zowe CLI on [z/OS Unix Systems Services](https://www.ibm.com/docs/en/zos/2.4.0?topic=descriptions-zos-unix-system-services) is not supported at this time. If you would like to use USS in the future, show your interest by voting for the enhancement in the [Zowe CLI GitHub repository](https://github.com/zowe/zowe-cli/issues/1680). If enough votes are cast, the enhancement is added to the Zowe CLI dev team backlog.
```
anaxceron, line 35: @adam-wolfe Let me know if this is good as a reference/link to the USS github issue

t1m0thyj, line 35: Suggested change: `If you would like to use USS` -> `If you would like to use it on USS`

t1m0thyj (commented):
Suggested minor wording changes, everything else LGTM 🙂

docs/user-guide/cli-using-usingcli.md
adam-wolfe: I think we could remove the last sentence. We don't really have a threshold for how many 👍 we are looking for - it would just be another way to gauge interest so we can better prioritize the work.

anaxceron: fixed -- thank you!

anaxceron: I think we do have a count -- at least according to the bot: 
image

but I'm removing the sentence regardless, thank u

adam-wolfe (approved):

t1m0thyj (approved):

gejohnston (approved):
Everything looks good.

PR #2810, opened 2023-04-13
doc for how to change log and temp paths

anaxceron:
Documenting dev work for [Issue 2186](https://github.com/zowe/vscode-extension-for-zowe/issues/2186) and [PR 2189](https://github.com/zowe/vscode-extension-for-zowe/pull/2189):

Updating [Troubleshooting Zowe Explorer](https://docs.zowe.org/stable/troubleshoot/ze/troubleshoot-ze) to include documentation about how to change the default paths for ZE `logs` and `temp` folders


docs/images/troubleshoot/ZE/new-temp-folder-path.PNG
t1m0thyj, line 1: Should this file be renamed to have a lowercase extension (`.png` instead of `.PNG`)? When reviewing this PR, the image is failing to load, presumably because the path is case sensitive:
image

I'm not sure if paths are case sensitive in the Netlify build, or only for GitHub PR reviews, but this seems like it would be an easy change to make 🙂

anaxceron: I changed the casing on the ext to all lowercase. I think that fixed it.

docs/troubleshoot/ze/troubleshoot-ze.md
```diff
@@ -40,4 +40,38 @@ To fix the configuration file causing the error, follow these steps:
 2. Modify the file as needed and save the changes.
 3. Reload Visual Studio Code to confirm that Zowe Explorer can read the updated file.
 
+## Missing write access to VS Code `extensions` folder
```
sbohnert, line 43: Line 30: Delete comma.
Line 32: Delete ", follow these steps"

docs/troubleshoot/ze/troubleshoot-ze.md
```diff
@@ -40,4 +40,38 @@ To fix the configuration file causing the error, follow these steps:
...
 
+In some environments, the path for VS Code extensions (typically `~/.vscode/extensions`) can be read-only. This can happen when an environment has developers share a common read-only extensions folder that is managed by an admin with write access.
+
+In these cases, Zowe Explorer fails to activate because it cannot write to the `logs` and `temp` folders in the extension path. When Zowe Explorer launches, an error message displays, as in the following examples.
```
sbohnert, line 47: Change ", as in the following examples" to ". See the following examples"

docs/troubleshoot/ze/troubleshoot-ze.md
```diff
@@ -40,4 +40,38 @@ To fix the configuration file causing the error, follow these steps:
...
+
+![Logs folder write access error](../../images/troubleshoot/ZE/write-access-error-logs-folder.png)
+
+`temp` folder write access error:
+
+![Logs folder write access error](../../images/troubleshoot/ZE/write-access-error-temp-folder.png)
+
+To avoid this, change the `logs` and `temp` folder locations.
+
+Follow these steps:
```
sbohnert, line 59: Delete

docs/troubleshoot/ze/troubleshoot-ze.md
```diff
@@ -40,4 +40,38 @@ To fix the configuration file causing the error, follow these steps:
...
+
+1. In Zowe Explorer, select the **File menu**, select **Preferences**, and click on **Settings**.
+
+2. In either the **User** or **Workspace** tab, click on the **Extensions** option to open the menu.
```
sbohnert, line 63: Any reason why we cannot list only one option? What is the most common method?

docs/troubleshoot/ze/troubleshoot-ze.md
```diff
@@ -40,4 +40,38 @@ To fix the configuration file causing the error, follow these steps:
...
+
+3. Select **Zowe Explorer**.
+
+4. Enter the new path in the **Logs Folder** or **Temporary Downloads Folder** fields, as in the following examples:
```
sbohnert, line 67: For example

docs/troubleshoot/ze/troubleshoot-ze.md
```diff
@@ -40,4 +40,38 @@ To fix the configuration file causing the error, follow these steps:
...
+
+`temp` folder write access error:
+
+![Logs folder write access error](../../images/troubleshoot/ZE/write-access-error-temp-folder.png)
+
+To avoid this, change the `logs` and `temp` folder locations.
+
+Follow these steps:
+
+1. In Zowe Explorer, select the **File menu**, select **Preferences**, and click on **Settings**.
```
adam-wolfe, line 61: I would change "In Zowe Explorer" to "In VS Code"

docs/troubleshoot/ze/troubleshoot-ze.md
t1m0thyj: VS Code allows settings to be defined across different scopes ([source](https://code.visualstudio.com/docs/getstarted/settings)):
> * User Settings - Settings that apply globally to any instance of VS Code you open.
> * Workspace Settings - Settings stored inside your workspace and only apply when the workspace is opened.

The settings for logs and temp folders could be changed in either place - either globally for all VS Code instances, or only for the current project/workspace. However I think we could omit this part: "in either the **User** or **Workspace** tab".

Either the user should know enough to select a scope without us telling them, or if they don't then setting it globally is probably what they want, and is the default behavior because the most common tab that's selected by default is the first one (User).

anaxceron: Leaving as it b/c the intention is to include the tabs as a wayfinder for the Extensions tab. It's not so much for the user to figure out which option they want, but it's meant as an indication of the area where they can find the Extensions tab.

docs/images/troubleshoot/ZE/new-temp-folder-path.PNG
t1m0thyj: The image is still failing to load on GitHub and shows up with an uppercase extension in the file tree:
image

anaxceron: this has been resolved

PR #2865, opened 2023-05-05
fix: Update `docusaurus` to 2.4.1; fix swizzled React components

traeok:

This PR updates Docusaurus to the latest version (2.4.1). In addition, it resolves any broken React components as a result of updating from 2.0.0-beta.3. I've also made a couple adjustments to those components so that they do not break as a result of missing data. I built a list of breaking changes that I ran into from 2.0.0-beta.3 to 2.4.1:

- `versionMetadata` has moved
	- **2.0.0-beta.3:** Prop in `DocItem` component
	- **2.4.1:** result of `useDocsVersion` in `@docusaurus/theme-common/internal` 
- `themeConfig`
	- `hideableSidebar: true`
		- **2.0.0-beta.3:** available option in `themeConfig`
		- **2.4.1:** _moved_ to `themeConfig {} -> docs {} -> sidebar {} -> hideable`
	- `googleAnalytics`
		- **2.0.0-beta.3:** available option in `themeConfig`
		- **2.4.1:** _moved_ to `@docusaurus/preset-classic` config
- `useAlgoliaContextualFacetFilters` import has changed
	- **2.0.0-beta.3:** Default export at `@theme/hooks/useAlgoliaContextualFacetFilters`
	- **2.4.1:** Named export at `@docusaurus/theme-search-algolia/client`
- `useVersions, useLatestVersion, useActiveDocContext` import has moved
	- **2.0.0-beta.3:** `@theme/hooks/useDocs`
	- **2.4.1:** `@docusaurus/plugin-content-docs/client`
- `InitialMDXComponents` import has moved
	- **2.0.0-beta.3:** `@theme-init/MDXComponents`
	- **2.4.1:** `@theme/MDXComponents`
- `useActiveVersion` import has moved
	- **2.0.0-beta.3:** `@theme/hooks/useDocs`
	- **2.4.1:** `@docusaurus/plugin-content-docs/client`
- `useTitleFormatter` import has moved
	- **2.0.0-beta.3:** `@docusaurus/theme-common`
	- **2.4.1:** `@docusaurus/theme-common/internal`
- `useActivePlugin` import has moved
	- **2.0.0-beta.3:** `@theme/hooks/useDocs`
	- **2.4.1:** `@docusaurus/plugin-content-docs/client`
- `useDocVersionSuggestions` import has moved
	- **2.0.0-beta.3:** `@theme/hooks/useDocs`
	- **2.4.1:** `@docusaurus/plugin-content-docs/client`
- `useTOCHighlight` import has moved
	- **2.0.0-beta.3:** `@theme/hooks/useTOCHighlight`
	- **2.4.1:** `@docusaurus/theme-common/internal`
	- _Note:_ This is now an internal function, probably shouldn't be used
- `colorMode.switchConfig` is deprecated
        - Note from docusaurus PR [6771](https://github.com/facebook/docusaurus/pull/6771): If you want to customize the icons for light and dark mode, swizzle `IconLightMode`, `IconDarkMode`, or `ColorModeToggle` instead

---

:heart:Thank you!


traeok:
The Netlify builds are still failing from time to time. This is likely due to RAM limitations. The site peaks around 8GB during asset processing and without disabling minification, it is difficult to avoid this much RAM usage.

Node.js has the v8 argument `--max-old-space-size` which defaults to 1400MB on 64-bit systems. If we set the value too high (e.g. 9126) the process will be killed by the system if it reaches that value. If the value is too low, Node.js garbage collection can fail and the process will end as a result.

The best option to reduce RAM usage here is to find the optimal value for `--max-old-space-size`. In addition, since minification should only affect initial loading for some users, I've added back the `--no-minify` arg in hopes that it will reduce the RAM footprint enough for Netlify to build successfully.

src/theme/TOC/index.js
```diff
@@ -2,25 +2,21 @@
...
 import styles from "./styles.module.css";
 const LINK_CLASS_NAME = "table-of-contents__link";
 const ACTIVE_LINK_CLASS_NAME = "table-of-contents__link--active";
 const TOP_OFFSET = 100;
 /* eslint-disable jsx-a11y/control-has-associated-label */
 
 function Headings({ toc, isChild }) {
-  if (!toc.length) {
-    return null;
-  }
```
t1m0thyj, line 15: Should we restore this check with `?` added? `if (!toc?.length) {`

This would keep the old behavior of the component to return null, and then we could remove `?` from the `toc.map` below.

traeok:
Going to table this for now, as although the builds now succeed, there are a couple discrepancies with the layout that was introduced as a part of updating to Docusaurus 2.4.0. Hopefully I should be able to address those soon, but for now #2868 will resolve the random build failures - thanks @t1m0thyj !

traeok:
Pushing up commit shortly with following fixes:

- Update to Docusaurus 2.4.1
- Fixed all bugs from updating to 2.4.1 w/ React components
- Fixed formatting for share buttons at top of documents
- Remove TOC component as it is **forbidden** to swizzle, according to Docusaurus
  - Replaced with wrapper around `TOCItem`, which should be easier to maintain moving forward
- Updated `DocItem` component so that it uses the new `DocProvider` recommended by Docusaurus
- Added dynamic TPSR link, so that TPSR links redirect to the version the user is browsing
- Fixed "Release Notes" link for latest version as it is currently hardcoded to v1.23 release notes
  - To reproduce on `main` branch: Click on version dropdown -> `All versions` -> click on "Release Notes" for the latest version
- Fixed alignment issues with Yes/No feedback buttons
- Fixed issue where markdown components had a circular include
- Fixed some CSS rules for better formatting

PR #2868, opened 2023-05-10
Use GitHub Actions to build docs site

t1m0thyj:
Fixes out-of-memory errors when building the docs site. Currently the docs site requires around 8 GB of memory to build.

So far we have been using Netlify build machines for our CI/CD - both to build the site with Docusaurus, and deploy it to production or a preview site. The problem is that Netlify's free plan relies on shared build servers with no guaranteed amount of memory. We get allocated an arbitrary amount between 3 and 11 GB, so many builds fail because they run out of memory.

This PR moves the first step of the process (building with Docusaurus, which is the memory intensive part) to run with GitHub Actions instead. As part of the Zowe org, we have access to GitHub build machines for free that do have a guaranteed amount of memory. There will always be 11 GB available (7 GB real + 6 GB swap - 2 GB overhead) which should be more than enough.


t1m0thyj:
Made some changes based on our discussion this morning 🙂 
* Link Checker workflow can be triggered manually to run on any branch. This change won't take effect until the PR is merged. When that happens, I can post a GIF showing how to kick it off manually.
* Netlify build/deploy workflow will run on all pull requests now to create a preview build. This also won't take effect until the PR is merged.

Immediately before merging this pull request, we should change this Netlify setting - otherwise we will have duplicate builds:
image

samanthasusu (approved):

PR #2882, opened 2023-05-17
fixing broken links

anaxceron:
Fixing broken links as a result of moving ZE content to the ZE wiki. Fixing other links found in Link Checker


zFernand0 (approved):
Thanks for fixing the links! 
LGTM! 😋

PR #2888, opened 2023-05-22
adding job cancellation to ZE sample use cases

anaxceron:
Added the **Cancelling a job** section to [Sample Use Cases](https://docs.zowe.org/stable/user-guide/ze-use-cases/)


traeok (approved):
LGTM, thanks @anaxceron!

PR #2892, opened 2023-05-23
ZE known limitation re: creds

anaxceron:
Added new section re: ZE known limitation surrounding credentials and virtual environments

### DIFF
--- a/docs/troubleshoot/ze/ze-known-limits.md
+++ b/docs/troubleshoot/ze/ze-known-limits.md
@@ -1,0 +1,26 @@
+# Known Zowe Explorer limits
+
+### Mismatched credentials when using Zowe Explorer and Zowe CLI
+
+#### Limitation
+
+Mismatching credentials can potentially lock you out from using Zowe CLI and Zowe Explorer in Visual Studio Code with either Windows Subsystem for Linux (WSL) or Remote Secure Shell (SSH).
+
+WSL allows developers to run a Linux environment on Windows without the need for a separate virtual machine or dual booting. When using Zowe Explorer with WSL, two parallel processes take place: VS Code runs on the Windows operating system, while code execution and tooling happen within the Linux environment. 
+
+With Remote SSH, the network protocol provides users with a secure connection to a remote computer. When using this protocol with Zowe Explorer, you can securely connect to a remote machine or a remote development environment. 
+
+Both WSL and Remote SSH provide more tools for mainframe developers, but they also have limitations when it comes to credentials.
+
+Authentication information used in Zowe Explorer &mdash; such as usernames and passwords, SSH keys, API Mediation Layer tokens &mdash; resides on the developer’s local machine, not the Linux environment or the remote server or virtual machine. This is because credential storage is managed by VS Code, which runs on Windows.
+
+Using the VS Code integrated terminal with virtual machines does not include access to the credentials stored by Zowe Explorer in the local machine. Zowe CLI, for example, is not able to retrieve credentials saved on a developer’s PC when accessing the mainframe. Instead it attempts to use any credentials stored in the virtual machine.
+
+This can lead to confusion and inconsistencies when authenticating and accessing resources.
+
+#### Solution
+
+It is crucial to ensure that credentials are carefully managed between the local machine and the remote server to maintain proper authentication. 
+
+To avoid any potential issues (such as being locked out) caused by credential mismatch or discrepancies, credentials in both local and virtual/remote machines must match.
+
--- a/docs/troubleshoot/ze/ze-known-limits.md
+++ b/docs/troubleshoot/ze/ze-known-limits.md
@@ -1,4 +1,4 @@
-# Known Zowe Explorer limits
+# Known Zowe Explorer limitations
 
 ### Mismatched credentials when using Zowe Explorer and Zowe CLI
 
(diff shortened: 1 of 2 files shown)


docs/troubleshoot/ze/ze-known-limits.md
```diff
@@ -0,0 +1,26 @@
+# Known Zowe Explorer limits
```
adam-wolfe, line 1: I think "Known Zowe Explorer limitations" is clearer.

docs/troubleshoot/ze/ze-known-limits.md
```diff
@@ -0,0 +1,26 @@
...
+
+Mismatching credentials can potentially lock you out from using Zowe CLI and Zowe Explorer in Visual Studio Code with either Windows Subsystem for Linux (WSL) or Remote Secure Shell (SSH).
+
+WSL allows developers to run a Linux environment on Windows without the need for a separate virtual machine or dual booting. When using Zowe Explorer with WSL, two parallel processes take place: VS Code runs on the Windows operating system, while code execution and tooling happen within the Linux environment. 
+
+With Remote SSH, the network protocol provides users with a secure connection to a remote computer. When using this protocol with Zowe Explorer, you can securely connect to a remote machine or a remote development environment. 
+
+Both WSL and Remote SSH provide more tools for mainframe developers, but they also have limitations when it comes to credentials.
+
+Authentication information used in Zowe Explorer &mdash; such as usernames and passwords, SSH keys, API Mediation Layer tokens &mdash; resides on the developer’s local machine, not the Linux environment or the remote server or virtual machine. This is because credential storage is managed by VS Code, which runs on Windows.
```
adam-wolfe, line 15: Comment regarding "This is because credential storage is managed by VS Code, which runs on Windows."

VS Code can be run under Linux and MacOS as well. I think the key point is that VS Code will load credentials from the local filesystem (or whatever filesystem contains the OS that hosts VS Code) rather than from a WSL VM or remote SSH connection that VS Code could be using. Maybe @zFernand0 can check me on how this should be worded.

docs/troubleshoot/ze/ze-known-limits.md
```diff
@@ -0,0 +1,26 @@
...
+
+Using the VS Code integrated terminal with virtual machines does not include access to the credentials stored by Zowe Explorer in the local machine. Zowe CLI, for example, is not able to retrieve credentials saved on a developer’s PC when accessing the mainframe. Instead it attempts to use any credentials stored in the virtual machine.
+
+This can lead to confusion and inconsistencies when authenticating and accessing resources.
+
+#### Solution
+
+It is crucial to ensure that credentials are carefully managed between the local machine and the remote server to maintain proper authentication. 
+
+To avoid any potential issues (such as being locked out) caused by credential mismatch or discrepancies, credentials in both local and virtual/remote machines must match.
```
adam-wolfe, line 25: @zFernand0, I don't know that this is something we would want to mention here (to avoid confusing people), but it seems like there could be situations where a user may want to access services using different credentials in a VM or via ssh for testing purposes or to access different resources.

Would it be safer to say "credentials in both ... _should_ match."? It is acceptable as-is, but if there is nothing stopping a user who wants to intentionally use different credentials via VM or SSH, we may not want to tell them that they cannot do it.

docs/troubleshoot/ze/ze-known-limits.md
anaxceron: good point, that was awkward wording. updated.

docs/troubleshoot/ze/ze-known-limits.md
```diff
@@ -0,0 +1,26 @@
...
+
+With Remote SSH, the network protocol provides users with a secure connection to a remote computer. When using this protocol with Zowe Explorer, you can securely connect to a remote machine or a remote development environment. 
+
+Both WSL and Remote SSH provide more tools for mainframe developers, but they also have limitations when it comes to credentials.
+
+Authentication information used in Zowe Explorer &mdash; such as usernames and passwords, SSH keys, API Mediation Layer tokens &mdash; resides on the developer’s local machine, not the Linux environment or the remote server or virtual machine. This is because credential storage is managed by VS Code, which runs on Windows.
+
+Using the VS Code integrated terminal with virtual machines does not include access to the credentials stored by Zowe Explorer in the local machine. Zowe CLI, for example, is not able to retrieve credentials saved on a developer’s PC when accessing the mainframe. Instead it attempts to use any credentials stored in the virtual machine.
```
sbohnert, line 17: Change last line to:
"Instead, Zowe CLI attempts to ..."

docs/troubleshoot/ze/ze-known-limits.md
zFernand0: You do have a good point.
I believe I suggested Windows thinking only about WSL(2).
But it's possible for someone on any flavor of Linux or MacOS to be running VSCode and connecting over remote SSH to another machine.

How about...
"This is because credential storage is managed by VS Code, which stores them using the host's Operating System Credential Manager"?

(or some other flavor) 😋

docs/troubleshoot/ze/ze-known-limits.md
zFernand0: That's fair. 👍🏽 
We can be more lenient (._.?) in the way we state it 😋

zFernand0 (commented):
I agree with Adam's comments and suggested something for the first one 😋 
Thank you both! 🥳

adam-wolfe (approved):
LGTM

PR #2910, opened 2023-06-06
Remove failed merge conflict markers + various small updates to App Framework (see description)

DivergentEuropeans:
- Removing remnants of failed merge conflict markers
- v1.26 and onwards has hotkey support, so I added this section
- upcoming Zowe release will have a minor hotkey change, so I added this
- v2.3 and onwards has a minor hotkey change for Editor, so I added that
- v2+ doesn't support Workshop User Browser App (V1) so removed it from tutorials section for V2
- clarifying comment to TAB / SHIFT + TAB tabbing
- clarifying comment on how using * can be used in regex logger logic, if the user doesn't know the name of component loggers they'd like to activate (I know it's in the example, but I've had customers not know so making it explicit)
- hotkey --> keyboard shortcut


anaxceron (changes requested):
Let's avoid the term "hotkey." It may not be clear to some of users. Best practice is to use "keyboard shortcut" instead.

anaxceron (approved):
Thanks for replacing the use of "hotkey"! Lookin' good.

PR #2912, opened 2023-06-06
Remove CCS prereq from Install with z/OSMF 

janan07:
The CCS requirement is not applicable for installing with z/OSMF


anaxceron (approved):
Thanks for updating all our older versions w/ the correct documentation!

PR #2924, opened 2023-06-12
Create config scenarios from zowe yaml

janan07:
This PR describes the 5 certificate configuration scenarios presented in the zowe yaml in more human readable format.
The original yaml file is: https://github.com/zowe/zowe-install-packaging/blob/60bc4b44ecd502fcec640fbb9e2874e9d56e826a/example-zowe.yaml#L142


# Certificate configuration scenarios.md
```diff
@@ -0,0 +1,246 @@
...
+      country: ""
+    ```
+9. Set the validity in days for the Zowe generated certificates
+    ```  
+    validity: 3650
+    ```
+10. Set the domain names and IPs specified in nested subsection `SAN`. If this field is not defined, the `zwe init` command uses the value `zowe.externalDomains`.
+    ```
+    san:
+        sample domain name
```
samanthasusu, line 63: `sample domain name` and `sample IP address` are commented in yaml. It would be better to put them after `#` or just remove them.

# Certificate configuration scenarios.md
```diff
@@ -0,0 +1,246 @@
...
+* [Scenario 3: Use a z/OS Keyring with Zowe generated certificates](#scenario-3-use-a-zos-keyring-with-zowe-generated-certificates)
+* [Scenario 4: Use a z/OS Keyring and connect to an existing certificate](#scenario-4-use-a-zos-keyring-and-connect-to-an-existing-certificate)
+* [Scenario 5: Use a z/OS Keyring and import a certificate stored in a data set](#scenario-5-use-a-zos-keyring-and-import-a-certificate-stored-in-a-data-set)
+
+## Scenario 1: Use a PKCS12 (keystore) with Zowe generated certificates
+
+1. Select the type of certificate storage with `type: PKCS12`  
+Possible values include the following values:  
+
+    * PKCS12
```
janan07, line 17: Is this list useful? If not, I suggest it be omitted.

# Certificate configuration scenarios.md
```diff
@@ -0,0 +1,246 @@
+# Certificate configuration scenarios
+
+There are five scenarios/use cases for configuring a certificates.
```
pinpan:
```suggestion
There are five scenarios/use cases for configuring certificates.
```

# Certificate configuration scenarios.md
```diff
@@ -0,0 +1,246 @@
...
+Use the applicable certificate configuration scenario according to your needs.
+
+* [Scenario 1: Use a PKCS12 (keystore) with Zowe generated certificates](#scenario-1-use-a-pkcs12-keystore-with-zowe-generated-certificates)
+* [Scenario 2: Use a PKCS12 (keystore) and import a certificate generated by another CA](#scenario-2-use-a-pkcs12-keystore-and-import-a-certificate-generated-by-another-ca)
+* [Scenario 3: Use a z/OS Keyring with Zowe generated certificates](#scenario-3-use-a-zos-keyring-with-zowe-generated-certificates)
+* [Scenario 4: Use a z/OS Keyring and connect to an existing certificate](#scenario-4-use-a-zos-keyring-and-connect-to-an-existing-certificate)
```
pinpan, line 9:
```suggestion
* [Scenario 4: Use a z/OS Keyring and connect an existing certificate](#scenario-4-use-a-zos-keyring-and-connect-to-an-existing-certificate)
```

# Certificate configuration scenarios.md
1000TurquoisePogs: Scenario 1 implies PKCS12. The rest of the list here would not be possible.

# Certificate configuration scenarios.md
```diff
@@ -0,0 +1,246 @@
...
+  **Example zowe yaml for scenario 1**  
+  
+**_ADD EXAMPLE_**
+
+## Scenario 2: Use a PKCS12 (keystore) and import a certificate generated by another CA
+
+1. Select the type of certificate storage with `type: PKCS12`  
+Possible values include the following values:    
+
+    * PKCS12
```
1000TurquoisePogs, line 78: Scenario 2 implies PKCS12. The rest of the list here would not be possible.

# Certificate configuration scenarios.md
```diff
@@ -0,0 +1,246 @@
...
+  **Example zowe yaml for scenario 2**  
+  
+**_ADD EXAMPLE_**
+
+##  Scenario 3: Use a z/OS Keyring with Zowe generated certificates
+
+1. Select the type of certificate storage with `type: JCERACFKS`  
+Possible values include the following values:  
+
+    * PKCS12
```
1000TurquoisePogs, line 125: Scenario 1 implies JCE*. PKCS12 is not possible here.

# Certificate configuration scenarios.md
```diff
@@ -0,0 +1,246 @@
...
+    ```
+    **Note:** PEM format certificate authorities can be imported and trusted.
+
+  **Example zowe yaml for scenario 2**  
+  
+**_ADD EXAMPLE_**
+
+##  Scenario 3: Use a z/OS Keyring with Zowe generated certificates
+
+1. Select the type of certificate storage with `type: JCERACFKS`
```
1000TurquoisePogs, line 122: Not type JCERACFKS, rather, one of several names listed below starting with JCE

# Certificate configuration scenarios.md
```diff
@@ -0,0 +1,246 @@
...
+    ```
+    **Note**: Due to the limitation of the `RACDCERT` command, this field should contain exactly 2 entries with the domain name and IP address.
+
+  **Example zowe yaml for scenario 3**  
+  
+**_ADD EXAMPLE_**
+
+## Scenario 4: Use a z/OS Keyring and connect to an existing certificate
+
+1. Select the type of certificate storage with `type: JCERACFKS`
```
1000TurquoisePogs, line 180: Same issue as above

# Certificate configuration scenarios.md
```diff
@@ -0,0 +1,246 @@
...
+        ```
+        **Note:** Due to the limitation of `RACDCERT` command, this field should contain a maximum of 2 entries.  
+
+  **Example zowe yaml for scenario 4**  
+  
+**_ADD EXAMPLE_**
+
+## Scenario 5: Use a z/OS Keyring and import a certificate stored in a data set
+
+1. Select the type of certificate storage with `type: JCERACFKS`
```
1000TurquoisePogs, line 215: Same issue as above.

docs/user-guide/api-mediation/certificate-configuration-scenarios.md
```diff
@@ -0,0 +1,252 @@
...
+This automation can be performed by defining and customizing the `zowe.setup.certificate` section of your Zowe YAML configuration.
+Zowe can then automate the certificate setup via the `zwe init certificate` command. This command is run as a subcommand of `zwe init`.
+
+**Note:**  
+Automated generation of certificates is an option, but is not required. If you already have a keystore that contains a valid certificate (see the description of a valid certificate), and the certificate's corresponding private key, along with a truststore which validates the certificate and any other certificates you expect to encounter, then you also have the option of directly defining the parameter `zowe.certificate` which specifies the location of each of these objects. Note that this parameter should not be confused with the parameter `zowe.setup.certificate`.
+
+## What is a valid certificate in Zowe?
+
+A valid certificate for use in Zowe can be either of the following formats:
```
1000TurquoisePogs, line 16: When you say "formats", I think file-based or keyring based.
The "formats" are as stated _above_. ("For use of Zowe on z/OS, keystore and truststore can either be unix file-based (PKCS12) or z/OS keyring-based.")
The _following_ bullets are what I'd think of as certificate attributes instead.
So you might say,

"A valid certificate for use in Zowe must have one of the following attributes, regardless of format:"

docs/user-guide/api-mediation/certificate-configuration-scenarios.md
```diff
@@ -0,0 +1,252 @@
+# Certificate configuration scenarios
+
+As a system programmer, review the different scenarios for configuring Zowe for automatic certificate setup. Examples are provided for each scenario.
+
+Zowe servers require both a keystore to store the certificates as well as a truststore to validate of these certificates and any other certificates that are encountered.
+For use of Zowe on z/OS, keystore and truststore can either be unix file-based (PKCS12) or z/OS keyring-based. 
+Both the keystore and truststore can automatically be created by Zowe regardless of which storage type is used. Keystores and truststores can be populated either with certificates that the user chooses, or with self-signed certificates generated by Zowe. 
+This automation can be performed by defining and customizing the `zowe.setup.certificate` section of your Zowe YAML configuration.
```
1000TurquoisePogs, line 8: I find it intuitive to use `foo.bar.baz` format to quickly describe which section of the YAML is being described.
But, I have to assume nothing with our users. Should we hyperlink the heck out of these code blocks once or every time they appear like so? [`zowe.setup.certificate`](https://github.com/zowe/zowe-install-packaging/blob/v2.x/staging/example-zowe.yaml#L99) [[schema]](https://github.com/zowe/zowe-install-packaging/blob/v2.x/staging/schemas/zowe-yaml-schema.json#L163) [[syntax]](using-yaml.md)

We need to make such a "using-yaml.md" page to link... but basically, those 3 links would in theory completely contextualize a section of config, but would be tedious to do every single time; we'd probably forget.

docs/user-guide/api-mediation/certificate-configuration-scenarios.md
janan07: Attributes sounds good.

docs/user-guide/api-mediation/certificate-configuration-scenarios.md
janan07: Does this mean linking to the yaml in all cases? I think a single link to the yaml would be sufficient, but also include the particular configuration per each scenario in the yaml example section. Perhaps Joe has these customized yaml examples as per scenario...

1000TurquoisePogs: example-zowe.yaml is currently our 1 and only YAML file. We're going to split it into several to allow the doc to more easily say "grab this whole file" or "see this whole file", but that's coming in a few weeks/months.

Almost all of the `foo.bar.baz` stuff can be pinpointed in example-zowe.yaml today, so we COULD link to it every single place we do this, but again I ask do we want to as a rule? I think our users would like it more than they dislike the clutter of doing that, but just my opinion...

Schema on the other hand, there really is only 1 for that purpose. Schema describes whats valid beyond examples. The example may show 1 thing, but the schema can tell you there's 5 alternatives.

As for "Syntax", I think we need a page much like this which tells users how to use YAML and what we mean by `this.that.other` https://www.ibm.com/docs/en/umsfz/1.2.0?topic=references-yaml-format

janan07: I think this would provide a high degree of detail and could serve users well for now until the "grab this whole file" solution is developed... hopefully sooner than later.

pinpan: My 2 cents ... I think we should follow one major approach how to help the users to understand our documents. 

One approach is to provide educational material, point the uninformed users there once from every dependent document, and later expect that they either learned the concepts or at least can navigate themselves to that place anytime they need. 

The other approach would be to not bet on the user's knowledge, and provide the links everywhere. 

The first one is IMHO cleaner and easier to maintain, while the later one would clutter the text with repetitive links without significant benefits for the user. I can imagine they'd open a new tab to the same document (yaml, schema,...) every time they see a link, or even worse, may open it in the same tab replacing the main document, and lose their focus.

docs/user-guide/api-mediation/certificate-configuration-scenarios.md
```diff
@@ -0,0 +1,252 @@
+# Certificate configuration scenarios
+
+As a system programmer, review the different scenarios for configuring Zowe for automatic certificate setup. Examples are provided for each scenario.
+
+Zowe servers require both a keystore to store the certificates as well as a truststore to validate of these certificates and any other certificates that are encountered.
```
pinpan, line 5:
```suggestion
Zowe servers require both a keystore to store the certificates as well as a truststore to validate these certificates and any other certificates that are encountered.
```

achmelo (approved):

Issue #2903, opened 2023-06-01
Add section to docs on how to determine the version of Zowe (including all component)


nkocsis:
## Is your request for enhancement related to a problem? Please describe.
Describe how to determine what level of Zowe a system is on.

## Describe the solution you'd like
This should show the steps/commands to do this

samanthasusu:
Thank you Nick for proposing this action. I'm wondering if any standards you have for users to determine the version. If there are some existing standards or conditions that we can follow, would you mind providing them here so that we could have some discussion? Thank you!

nkocsis:
@Joe-Winchester Joe, is this something you can help with ?

anaxceron:
@janan07 @samanthasusu @nellietan -- I have added my client-side content to [Checking your Zowe version release number](https://docs.zowe.org/stable/troubleshoot/troubleshoot-check-your-zowe-version). Please add the content for your server-side components so that we can close out this issue.

anaxceron:
Closed as others have added/reviewed content.

PR #2937, opened 2023-06-14
first draft of finding-versions content

anaxceron:
First draft of how-to-find-component-version content. Addresses [Issue 2903](https://github.com/zowe/docs-site/issues/2903).


1000TurquoisePogs (commented):
This is great, makes sense to me.

2 other places to consider:

The zowe desktop login page and the gateway status page.
Those versions are slightly different - they're the build versions of the code, so instead of v2.10.0 it might say something like v2.10.0+20230705 or some other jibberish. It's not as friendly, but it's easy to find.

I'd also note a warning to the user:
It's possible to have forgotten to update your client components at the same time as your server components. When you have an issue, you want to make sure they're on the same versions!

anaxceron:
> Thanks @1000TurquoisePogs  -- I really appreciate the input

I added a user warning note, please take a look if you get a chance.

Re: the two other places: Since these are server-side components, I'll leave this documentation to @samanthasusu and @nellietan. The plan w/ this article was to start w/ client-side content, then hand this off to the other writers to add their documentation. The only reason I included the server side content that I did was b/c it already existed in another article ([Understanding the Zowe release](https://docs.zowe.org/stable/troubleshoot/troubleshoot-zowe-release)) and I wanted to re-organize the content for the sake of cohesion.

docs/troubleshoot/check-your-zowe-version.md
```diff
@@ -0,0 +1,115 @@
...
+Using the `trace` mode:
+```
+zwe version --trace
+Zowe v2.9.0
+build and hash: v2.x/master#3248 (77fe6c5277e52474ebe9258c1b745949537dc398)
+Zowe directory: /SYSTEM/zowe/runtime
+```
+The `trace` mode  shows the location where the convenience build was extracted (such as `<RUNTIME_DIR>`). Run this when you want to confirm the location of your Zowe runtime directory.
+
+## Client side commands
```
zFernand0, line 60: Curious if the previous topic ("Using the Manifest file") should mention Server-side somewhere in the title?

zFernand0 (approved):
The content LGTM! :yum:

docs/troubleshoot/check-your-zowe-version.md
anaxceron: @zFernand0 Is the manifest file only applicable for the server side? I didn't realize that. I thought it could be used for both client and server use cases, no?

anaxceron: Yes, we should mention the client-side stuff under the **Using the manifest file** section. I'll reach out to you today about this.

anaxceron: After our discussion, moving the **Using the manifest file** section to under the **Server side** section.

zFernand0 (approved):

PR #2944, opened 2023-06-20
remove duplicate file ref of Configure the Zowe APIs

janan07:
This PR removed the duplicate reference of Configuring the Zowe APIs in v2.6
image


anaxceron (approved):
Thanks for taking care of this!

PR #2990, opened 2023-07-12
Revert "archiving v2.9.x"

anaxceron:
Reverts zowe/docs-site#2988 - content mistakenly merged to master when it should've been merged to docs-staging


zFernand0 (approved):
Discussed with Ana offline and it makes sense to revert this action on master 
image

PR #2994, opened 2023-07-12
updating style guide

anaxceron:
Made the following changes:
- Added "hotkeys" to the list of words to avoid
- Added listing on using punctuation in lists
- Added listing on release notes
- Other minor updates


samanthasusu (approved):
Looks good! Sorry, I looked at it before but forgot to approve it. Unforgivable again.

PR #3001, opened 2023-07-14
Increase zowe size requirement statement

1000TurquoisePogs:
Up for debate but here's my PR.
The requirements for zowe state you need a minimum of 833mb for zFS to fit Zowe on.

Extracted, zowe is about 510 MB right now, so fine.
But, I believe this number was supposed to include the archive + the extracted size, because without more than double the archive space, you can't extract Zowe.

That'd put us above 1000 MB as a minimum these days.
I think it's best to give growing room, and set this number as 1500 MB.

Also hate to be that guy but "mb" is a fictional unit of measurement "millibits", and I'm always surprised by the ways our users get confused, so let's avoid the issue.

### DIFF
--- a/docs/user-guide/systemrequirements-zos.md
+++ b/docs/user-guide/systemrequirements-zos.md
@@ -32,7 +32,7 @@ Be sure your z/OS system meets the following prerequisites:
 
    **Note:** z/OS V2.2 reached end of support on 30 September 2020. For more information, see the z/OS v2.2 lifecycle details [https://www.ibm.com/support/lifecycle/details?q45=Z497063S01245B61](https://www.ibm.com/support/lifecycle/details?q45=Z497063S01245B61). 
 
-- zFS volume has at least 833 mb of free space for Zowe server components, their keystore, instance configuration files and logs, and third-party plug-ins.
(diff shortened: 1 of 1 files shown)


DivergentEuropeans:
I agree w/ mb nitpick 100%

I'll defer to what I see in gaming industry space requirements ie. space requirement is usually the minimum space required, no DLC's (so our "Zowe core") w/ additional download artifacts removed. But then we say "and third-party plug-ins" I see no reason to put 1500 MB if we can say Zowe is under 1000 MB. But the archive size is still important, so why don't we just say like

"zFS volume needs at least 1000 MB of free space for core Zowe server components, their keystore, instance configuration files and logs, and core plug-ins (excluding the archive which can be safely removed after Zowe is installed).
However, 1500 MB is recommended for users to leave room for third-party plugins & future Zowe components and extensions"

1000TurquoisePogs:
I appreciate relating this to other spaces for comparison, but the comparison to the gaming industry isn't valid here.

I know that z/OS admins can be extremely conservative with USS partition space. They may interpret 833MB literally, and then end up with Zowe not working because you cannot simultaneously have an archive and extract it without having at least twice the space of the archive present. Its entirely possible the admins will make a 833MB (or even 1000MB, still not enough) partition, and then complain that they cant extract Zowe.

PR #3009, opened 2023-07-19
fixing incorrect commands

anaxceron:
Fixing incorrect commands in [Integrating with API Mediation Layer](https://docs.zowe.org/stable/user-guide/cli-using-integrating-apiml)


zFernand0 (approved):
thanks for fixing.
LGTM! 😋

PR #3017, opened 2023-07-20
adding new env variable

anaxceron:
Adding a new entry for the new ZOWE_V3_ERR_FORMAT env. variable


gejohnston (changes requested):
I suggest that we also state that this will be the only error format available in Zowe V3. Zowe will no longer use this environment variable in Zowe V3.

anaxceron:
@gejohnston I've updated the doc per your comment re: Zowe v3. Please take a look.

gejohnston (approved):
Looks good.

PR #3022, opened 2023-07-21
Add API ML feature to 2.10 release notes

janan07:
Last minute addition of a new feature in API ML for 2.10.


anaxceron (approved):
Looks good!

PR #3024, opened 2023-07-21
Update Configure certs in Installation checklists and header fixes

janan07:
This PR updates the content in the installation checklists with the current configure certificates articles and transforms some headers into gerunds as per the style guide


IgorCATech (approved):
LGTM

PR #3048, opened 2023-08-03
index test

anaxceron:
Turning ToC headers into pages with content


sidebars.js
```diff
@@ -37,6 +37,7 @@ module.exports = {
     {
       type: "category",
       label: "Release notes",
+      link: {type:"doc", id:"getting-started/release-notes"},
```
traeok, line 40: Try renaming `index.md` in the `release-notes` folder back to `release-notes-overview.md` and then update the `link` attribute like so:

```suggestion
      link: {type:"doc", id:"getting-started/release-notes/release-notes-overview"},
```

sidebars.js
```diff
@@ -37,6 +37,7 @@ module.exports = {
...
       items: [
         "getting-started/release-notes/release-notes-overview",
```
traeok, line 42: And then try removing this from the list of items so it doesn't show up twice under the "Release notes" category on the sidebar:

```suggestion
```

PR #3060, opened 2023-08-16
Remove v1 reference

### LINKED ISSUE #3019: Outdated v1 reference in docs.zowe.org/stable/user-guide/api-mediation/api-mediation-internal-configuration/
Within https://docs.zowe.org/stable/user-guide/api-mediation/api-mediation-internal-configuration/#environment-variables I see the section that says:

```
    Open the file <Zowe instance directory>/instance.env.

    Add a new line with the following property:

    apiml.security.auth.provider=saf. 
```

But, instance.env does not exist in v2.
It looks like that property suggested exists in zowe.yaml in v2, so that fix could be simple, but noting this in a ticket in case this page needs further review under the lense of if it contains v1 content that needs updating to v2.

balhar-jakub:
Replace v1 reference for environment variables with v2 method
 - https://github.com/zowe/docs-site/issues/3019

close #3019 

:heart:Thank you!

### DIFF
--- a/docs/user-guide/api-mediation/api-mediation-internal-configuration.md
+++ b/docs/user-guide/api-mediation/api-mediation-internal-configuration.md
@@ -116,10 +116,13 @@ One use case for adding an environmental variable is to change the authenticatio
 
 **Follow the steps:**
 
-1. Open the file `<Zowe instance directory>/instance.env`.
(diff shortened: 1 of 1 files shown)


janan07 (approved):
Reviewed by doc squad

PR #3072, opened 2023-08-21
App server

Martin-Zeithaml:
#### app-return-codes.md
Capital letter to be consistent:
 image
#### appserver-error-codes.md
- Missing subchapter for warning and error messages
- One message missing severity suffix
#### zis-error-codes.md
Changed header to be consistent with ZSS
image


janan07 (approved):
Reviewed by doc squad

PR #3080, opened 2023-08-29
Anax v2.11 branchsync docs

anaxceron:
Synching `master` and `docs-staging` branches in preparation for Zowe v2.11 doc.


anaxceron:
Banner resized for smart phone screens:

image

PR #3081, opened 2023-08-30
Refactor questionnaire

janan07:
This PR offers an important tip to consult with your security admin if configuring certificates for production and removes numerous links previously included which pertain to each question in the flow.,


balhar-jakub (commented):
It looks good to me

PR #3103, opened 2023-09-12
Update PDF creator

1000TurquoisePogs:
The PDF creator today is using an unmaintained library. I'm switching to a maintained fork of it, which the docusaurus website now recommends over the old one.

While trying this out, I noticed that I could also make the PDF easier to read by decreasing the font size via a css property.

I made another build target called docs:pdf:user which you do not need to run yet but should consider.
It does the same stuff as docs:pdf, but omits sections of the documentation that I thought were not relevant to users, such as developer content.

People should consider whether we should split the PDF into multiple in this way, and how many splits and which content go within each.


anaxceron (approved):
Looking forward to seeing the new PDF generator at work -- thanks for updating!

PR #3138, opened 2023-10-04
adding note re: non-ASCII characters and FTP

anaxceron:
Adding notes to [IBM z/OS FTP Plug-in for Zowe CLI](https://docs.zowe.org/stable/user-guide/cli-ftpplugin) and [Using Zowe Explorer FTP Extension](https://docs.zowe.org/stable/user-guide/ze-ftp-using-ze-ftp-ext) re: non-ASCII characters

### DIFF
--- a/docs/user-guide/cli-ftpplugin.md
+++ b/docs/user-guide/cli-ftpplugin.md
@@ -13,8 +13,10 @@ As a z/OS user, you can use the plug-in to perform the following tasks:
 
 ## Commands
 
-:::caution
+:::
+
 When transferring files, data sets, or data set members, use only ASCII characters. If a file contains non-ASCII characters (such as glyphs or mathematical symbols), a translation error can happen when the file is downloaded from, or uploaded to, the mainframe. This error can result in data loss.
+
 :::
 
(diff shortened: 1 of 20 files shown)


docs/user-guide/cli-ftpplugin.md
```diff
@@ -13,6 +13,10 @@ As a z/OS user, you can use the plug-in to perform the following tasks:
 
 ## Commands
 
+:::note
+When issuing commands, use only ASCII characters. FTP does not support non-ASCII characters, such as glyphs or mathematical symbols.
```
awharn, line 17: The issue in this case is not that the command does not contain ASCII characters. The issue in this case is that the file/dataset/dataset member being transferred contained characters that are not contained in the ISO-8859-1 (Latin-1/Extended ASCII) character set. When this occurs, a translation error character takes the place of the character that could not be represented, and if uploaded back to the mainframe, will result in a translation error and data loss.

awharn (approved):

docs/user-guide/cli-ftpplugin.md
anaxceron: thank you! updated based on your comments

adam-wolfe (approved):
LGTM

PR #3142, opened 2023-10-06
refactor drop down menu and some top-level article titles

janan07:
This PR refactors the drop down menu and standardizes titles in top-level articles so they have consistent naming conventions.


anaxceron (approved):
Looks good!

PR #3217, opened 2023-10-30
Language refactor of Installation overview

janan07:
This PR is to improve the readability of the Installation overview and to list the zwe init subcommands in the order they are performed in zwe init.


samanthasusu (approved):
looks good to me!

PR #3218, opened 2023-10-30
Reorganize content in Preparing for Installation

janan07:
This PR reorganizes the content in the article Preparing for Installation to improve readability.

I suggest we also create an article entitled "Creating a Zowe runtime directory" as I do not believe we have any procedure describing this.


samanthasusu (approved):
Pretty good enhancements!

PR #3239, opened 2023-11-08
Guide for enabling tracing on app framework

1000TurquoisePogs:
A user was asking how to enable debugging for app framework, and I didn't think the docs described it well or in the right places.
So, I created a new file in the troubleshooting section and dumped what I knew into it, referencing other pages.

appfw-tracing.md


docs/troubleshoot/app-framework/appfw-tracing.md
```diff
@@ -0,0 +1,49 @@
...
+Within the Zowe YAML file, the value `components.app-server.debug` can be set to `true` to turn on several debug loggers.
+This does not turn on every type of debugging but provides a basic set for debugging for the App Server.
+
+Enabling `components.app-server.debug` is equivalent to setting:
+
+```yaml
+components:
+  app-server:
+    logLevels:
+    - _zsf.*: 5
```
nosrednayduj, line 17: This is array syntax, and it should be object syntax, so it should not have that leading hyphen.

anaxceron: @nosrednayduj -- I have updated the file. Can you check that the syntax is correct?

PR #3240, opened 2023-11-08
Update AppFw configuration guide for extenders

1000TurquoisePogs:
The App Framework configuration guide for extenders became a big outdated around 2.10, when a larger emphasis was placed upon just using the Zowe YAML config file.

This made parts of the doc wrong, but the current state is actually simpler, so I removed more than I added in this PR.


anaxceron:
@1000TurquoisePogs 

- Please add the `mvd-server-config.md` file to the `sidebars.js` file. Right now this page isn't included in Zowe Docs.
- Also, I removed the ToC in `mvd-server-config.md` b/c a ToC will automatically display on the page's top right corner.

1000TurquoisePogs:
Okay done.

docs/extend/extend-desktop/mvd-server-config.md
```diff
@@ -1,15 +1,24 @@
 The Zowe's App Server and ZSS rely on many required or optional parameters to run, which includes setting up networking, deployment directories, plugin locations, and more.
```
janan07, line 1: This article does not conform to Zowe doc article standards. There should be one a single H1 header at the beginning of the article. Currently, there are multiple H1 headers and the title appears as the URL: mvd-server-config.

janan07 (approved):

PR #3271, opened 2023-11-24
Change the use case based configuration

balhar-jakub:
Signed-off-by: Jakub Balhar <jakub@balhar.net>

Proposed changes to the Advanced Gateway Configuration


janan07 (approved):
Reviewed and edited by Doc Squad

PR #3313, opened 2023-12-14
Setting latest_version to 2.13.0 and displaying getting_started on the sidebar

MarkAckert:
Brings docs live and corrects tags to display current version.


anaxceron (approved):
All it correct -- great catch, thank you Mark!!

PR #3329, opened 2023-12-22
Use Docusaurus sitemap plugin to remove old versions from SEO

t1m0thyj:
Currently we have a problem where Google indexes a lot of docs site content for old Zowe versions. This PR updates the sitemap generated by Docusaurus so that it only contains links for the latest version. The sitemap is crawled by Google to get a list of all the pages on our site, so hopefully pruning the sitemap will help.

For comparison, here is the live sitemap: https://docs.zowe.org/sitemap.xml
If you search the page for an old version like "2.7.x" there are hundreds of occurrences.

You can see the new sitemap on the deploy preview build for this PR: https://6585e5fc1eb6070c1cc8b4c3--zowe-docs-master.netlify.app/sitemap.xml
It only points to "/stable" and all the old version links have been removed.


anaxceron (approved):
Great idea -- hope this change takes care of things.

PR #3335, opened 2024-01-02
Auto-detect RC number in update typedoc script

t1m0thyj:
This PR enhances `scripts/updateTypedoc.sh` so that it auto-detects the latest RC number. Previously it had an RC number hardcoded and would fail if no bundle existed for that exact RC.

**How to Test**
Run the update typedoc script for Zowe 2.13.0 to verify that it downloads and extracts the RC2 zip file successfully.

### DIFF
--- a/scripts/updateTypedoc.sh
+++ b/scripts/updateTypedoc.sh
@@ -20,7 +20,8 @@ rm -rf typedoc
 rm -f zowe-nodejs-sdk-typedoc.zip
 
 # Download and extract new files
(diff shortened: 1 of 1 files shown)


zFernand0 (approved):
LGTM! 😋 
Took me a bit to test it 😅

anaxceron (approved):
Thank you for taking care of this @t1m0thyj -- this will definitely reduce the number of times I panic run to @zFernand0 for help

Issue #3367, opened 2024-01-17
ZWES1200W missing from docs.zowe.org/stable/troubleshoot/app-framework/zss-error-codes/


kevsenior:
## Description
In Zowe 2.13, I tried to create a new member in a PDS using the Editor app and received this error:
ZWES1200W Could not create file '/A506545.TEST.JCL/TEST1': Ret='129', res='93585469'
The message ZWES1200W is not documented.

## Pages to Update
docs.zowe.org/stable/troubleshoot/app-framework/zss-error-codes/

## Expected behavior
Documentation for ZWES1200W which contains pointers to where the Ret and Res codes are documented.

Martin-Zeithaml:
The doc change is in progress, there are more messages [missing](https://github.com/zowe/zss/issues/624).

PR #3390, opened 2024-01-22
ZSS Error Codes update

Martin-Zeithaml:
Added missing error codes for ZSS.

Fixes:
* https://github.com/zowe/docs-site/issues/3367
* https://github.com/zowe/zss/issues/624


Martin-Zeithaml:
I revert changes for ZWES1037E and ZWES1065E:

```
### ZWES1037E

  This is usually because the server port '%d' is occupied. Is ZSS running twice?

  **Reason:**

  What and why...

  **Action:**

  How to avoid this...

```

* The first line is identical to the message definition and this is what the customer will see in the log.
  * In this specific case '%d' will be replaced by a number, but the text will be the same.
  * This line should not be changed. If customer searches for text of the message, this will increase chance to be found.
  * If the message is too wrong (from the English point of view), it has to be changed in the source code too.
* **Reason** and **Action** are "free form".

janan07 (approved):
Reviewed by doc squad

PR #3443, opened 2024-02-20
Remove reference to "@include" in yaml doc

1000TurquoisePogs:
"@include" was never used, and never worked with configmgr, if it ever worked at all, yet we tell users they can use it.


lparadkar-rocket (approved):
Discussed with Sean and it can be merged.

PR #3454, opened 2024-02-21
Remove swap space in "Build docs site" workflow

t1m0thyj:
Describe your pull request here: Previously our Docusaurus builds were running out of memory, but now they are running out of disk space. Since GitHub Actions increased the amount of memory available for `ubuntu-latest` runners, we shouldn't need to allocate swap space anymore 😋 

List the file(s) included in this PR: builds-docs.yml workflow


traeok (approved):
LGTM, thanks @t1m0thyj

PR #2950, opened 2023-06-26
Improve Workflow to Generate List of Least Most Recently Modified Files

timgerstel:
### Your checklist for this pull request
🚨Please review the [guidelines for contributing](../docs/contribute/contributing.md) to this repository.

- [ ] If the changes in this PR is part of the next future release, make this pull request against the **docs-staging** branch which will be published at the next release boundary. If the changes in this PR are part of the current release, use the default base branch, **master**. For more information about branches, see https://github.com/zowe/docs-site/tree/master#understanding-the-doc-branches. 
      
- [ ] If this PR relates to GitHub issues in `docs-site` or other repositories, please list in Description, prefixed with **close**, **fix** or **resolve** keywords.

### Description (including links to related git issues)
Please describe your pull request.

:heart:Thank you!

PR #3571, opened 2024-03-26
Improve Workflow to Generate List of Least Most Recently Modified Files

1000TurquoisePogs:
See PR https://github.com/zowe/docs-site/pull/2950 for description. This is the same content but not on a fork.

Files added: See the files changed tab.
Note: Your "files added" automation is impossible to pass with the file type I have added.
It only checks for things.md or things.png but not things.yml
I wonder if this will now pass your files added checker because I have mentioned imaginary files that end with those?
Edit: no, it appears to be broken for all recent PRs.


lparadkar-rocket (approved):
approved

PR #3668, opened 2024-05-17
global AT-TLS configuration

achmelo:
Restructure the AT-TLS configuration guide so there is 1 general page that links to component-specific configuration.


docs/user-guide/api-mediation/configuration-at-tls.md
```diff
@@ -24,43 +23,12 @@ Review this article for descriptions of the configuration parameters required to
...
-```
-
 While API ML does not handle TLS on its own with AT-TLS enabled, API ML requires information about the server certificate that is defined in the AT-TLS rule. Esure that the server certificates provided by the AT-TLS layer are trusted in the configured Zowe keyring. Ideally, AT-TLS should be configured with the same Zowe keyring.
 
-2. If there is an outbound AT-TLS rule configured for the link between the API Gateway and z/OSMF, set the `zowe.zOSMF.scheme` property to `http`.
+If there is an outbound AT-TLS rule configured for the link between the API Gateway and z/OSMF, set the `zowe.zOSMF.scheme` property to `http`.
 
 :::note Notes
-* Currently, AT-TLS is not supported in the API Cloud Gateway Mediation Layer component.
+* AT-TLS is supported in the API Cloud Gateway Mediation Layer component since version 2.17.
```
pj892031, line 31: Actually, I guess 2.16 (without the new configuration of course) - see the https://github.com/zowe/common-java/pull/25

docs/user-guide/at-tls-configuration.md
```diff
@@ -0,0 +1,32 @@
...
+```yaml
+zowe:
+    network:
+        server:
+            attls: true
+```
+
+### High Availability Setup
+
+In a high availability setup where one LPAR (e.g., lpar1) is using AT-TLS and the second (e.g., lpar2) is also using AT-TLS, configure as follows:
```
pj892031, line 17: Just in case there is missing AT-TLS on any lpar?

achmelo: that's typo,thanks

balhar-jakub:
Doesn't this belong to the V3 documentation?

achmelo:
> Doesn't this belong to the V3 documentation?

is there some reason why it should not be in v2 as well?

balhar-jakub:
If it works for V2 as well, then there is probably none.

janan07 (approved):
Reviewed nd edited by doc squad

PR #3670, opened 2024-05-20
Fix APIML ATTLS documentation

1000TurquoisePogs:
Current documentation has several problems that will break if tried as-is

* Non-default ports (10* instead of 7551-7555)
* Port ranges that cover more ports than quantity of servers (10 ports instead of 5)
* Jobnames that catch unrelated parts of Zowe (ZWE*)
* Jobnames that don't catch any part of Zowe by default (ZWEA* instead of ZWE1*)
* Double "server:" entry in the gateway YAML
* Example only covering gateway & discovery - should probably list all servers to prevent confusion.

### DIFF
--- a/docs/user-guide/api-mediation/configuration-at-tls.md
+++ b/docs/user-guide/api-mediation/configuration-at-tls.md
@@ -41,12 +41,31 @@ zowe:
       server:
         ssl:
           enabled: false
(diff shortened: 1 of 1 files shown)


janan07 (approved):
reviewed by doc squad.

PR #3674, opened 2024-05-21
Add node v20 support to the documentation

1000TurquoisePogs:
Adding two things to the documentation:

* Support for node v20. This should work on older versions of Zowe but I'm not sure exactly how far back, so I am not updating the older version pages.
* Detail that nodejs is not required under specific circumstances provided you are running zowe of at least v2.16.0


janan07 (approved):
Reviewed by doc squad

PR #3676, opened 2024-05-22
Table for skipping steps within the install wizard

1000TurquoisePogs:
It is expected that users of the install wizard cannot do all operations by themselves, and hence there is a 'skip' option built in.
When involving another administrator, typically they have their own way of doing things and would not use the wizard in tandem. 
So, I constructed a table linking each install wizard step to an action and example, so that such a scenario can be covered via documentation.


janan07 (approved):
Reviewed by Doc squad

PR #3678, opened 2024-05-23
Update ZWES1606W with specific case

Martin-Zeithaml:

Specific case for warning message `ZWES1606W`, which could possibly help users.

*  docs/troubleshoot/app-framework/zss-error-codes.md


1000TurquoisePogs:
1606W has so many causes. It is our generic "we cannot reach the mediation layer"

The mediation layer could be down. 
It could be that you're using TLS with AT-TLS causing double TLS
It could be a keyring that has bad CAs that fail cert verification
DVIPA causing different IPs than expected

Martin-Zeithaml:
Yes, but I still believe that better one possible explanation than none. And recent history indicates this one is quite popular.

After this update I am expecting zowe people will start the avalanche of other possible explanations...

balhar-jakub:
Can we going forward try to differentiate these causes?

What would that take from the API ML side?

Martin-Zeithaml:
Yes, we can - but my experience is, what is not done during the development is lost in tommorowland.
So my personal preference is to close this PR as it is.
Each time, anyone will observe undocumented combination of RC/RSN and have possible explanation, this message can be updated.

1000TurquoisePogs:
I agree with @Martin-Zeithaml we shouldn't let this linger and will return in future PRs.


For ideas we can work on in some other tickets:

I don't think APIML can make this error much better.
Its a client error - ZSS is trying to be a client of gateway, and for various reasons the TLS communication can fail.
For example, if certificate verification fails, gateway would not know that.

One thing we can do better is to check certificates early. Zowe just starts up and either succeeds or fails but its often not clear why. I know we have tools for this, but they are not part of the startup, so people only learn their certificates are invalid after debugging.

1000TurquoisePogs (approved):

PR #3685, opened 2024-05-28
Native TLS documentation

1000TurquoisePogs:
Since zowe v2.13, it's been possible to configure some servers tls connections via `zowe.network.server`
But in v3, it's pretty much ALL the servers that support this now, so I've made v3 doc for this ability.
 [source](https://github.com/zowe/zowe-install-packaging/blob/v2.13.0/schemas/zowe-yaml-schema.json#L901)

### DIFF
--- a/docs/appendix/zwe_server_command_reference/zwe/zwe-install.md
+++ b/docs/appendix/zwe_server_command_reference/zwe/zwe-install.md
@@ -8,11 +8,13 @@
 
 After you extract Zowe convenience build, you can run this command to install
 MVS data sets.
+After installation of Zowe via SMP/E, PSWI, or convenience build, run this command to install sample modules and some load modules into MVS data sets.
 
 If you are using SMPE build, you can skip this command since MVS data sets are
 already prepared during SMPE install.
 
 These Zowe YAML configurations showing with sample values are used:
+The following Zowe YAML configuration example uses sample values:
 
 ```yaml
 zowe:
@@ -22,14 +24,25 @@ zowe:
 ```
 
 Expected outputs:
+**Expected outputs:**
 
 - Will create these data sets under `zowe.setup.dataset.prefix` definition:
   * `SZWEAUTH` contains few Zowe load modules (++PROGRAM).
   * `SZWESAMP` contains several sample configurations.
   * `SZWEEXEC` contains few utilities used by Zowe.
   * `SZWELOAD` contains config manager for REXX.
+The following data sets are created under `zowe.setup.dataset.prefix` definition:
+  * **SZWEAUTH**   
+    Contains a few Zowe load modules (++PROGRAM).
+  * **SZWESAMP**  
+   Contains several sample configurations.
+  * **SZWEEXEC**  
+   Contains a few utilities used by Zowe.
+  * **SZWELOAD**   
+   Contains config manager for REXX.
 
 ## Examples
+**Example:**
 
 ```
 zwe install -v -c /path/to/zowe.yaml
--- a/docs/user-guide/configure-zowe-zosmf-workflow.md
+++ b/docs/user-guide/configure-zowe-zosmf-workflow.md
@@ -99,9 +99,9 @@ The **Create configuration** step is mandatory. A valid zowe.yaml is required to
 
 This step consumes the zowe.yaml configuration file you created in the previous step. It contains three sub-steps:
 
-1. **(Optional) Run the Zowe install (only for convenience build)**. This step is required only for the convenience build. It runs the `zwe install` command.
-2. **Run the Zowe init** runs `zwe init mvs` and `zwe init stc` commands.
-3. **Run the Zowe init security** is optional if you have already previously preformed security setup for Zowe. This sub-step requires Security Administrator permissions. It runs the `zwe init apfauth` and `zwe init security` commands. If security is not set up for Zowe, contact your Security Administrator to perform this setup.
+1. **Run the Zowe install** to run the `zwe install` command.
+2. **Run the Zowe init** to run `zwe init mvs` and `zwe init stc` commands.
+3. **Run the Zowe init security** is optional if you have already previously preformed security setup for Zowe. This sub-step requires Security Administrator permissions. This runs the `zwe init apfauth` and `zwe init security` commands. If security is not set up for Zowe, contact your Security Administrator to perform this setup.
 
 :::note
 An alternative to executing **Run the Zowe init security** is to submit `ZWESECUR` JCL. For details, see [Configuring security](configuring-security.md).


docs/user-guide/tls-configuration.md
```diff
@@ -0,0 +1,85 @@
+# Customizing Native TLS
+
+Zowe's servers have built-in TLS support to enable HTTPS connections.
```
anaxceron, line 3: If you want to break up these sentences into two paragraphs, you need a blank line here to create a line break.

docs/user-guide/tls-configuration.md
```diff
@@ -0,0 +1,85 @@
...
+This is the default, and an alternative to using AT-TLS which is [documented here](./at-tls-configuration).
+
+:::info Required roles: security administrator
+:::
+
+## Server Parameters
+
+Each Zowe server can be customized either by defining attributes within the `zowe.network.server` object of the Zowe YAML configuration file. The same object can be put within an individual component's configuration, such as `components.zss.zowe.network.server` for ZSS, which will allow you to customize each component separate from others.
+Extensions are recommended to adhere to this configuration, but you must check with documentation for extensions to be sure.
```
anaxceron, line 12: I think you need a blank line here to create a line break.

docs/user-guide/tls-configuration.md
```diff
@@ -0,0 +1,85 @@
...
+      network:
+        server:
+          listenAddresses:
+          - 2.3.4.5
+```
+
+
+### TLS Versions
+
+By default, Zowe servers use TLSv1.3.
```
anaxceron, line 40: I think you need a blank line here to create a line break.

docs/user-guide/tls-configuration.md
```diff
@@ -0,0 +1,85 @@
...
+zowe:
+  network:
+    tls:
+      minTls: "TLSv1.2"
+      maxTls: "TLSv1.3"
+```
+
+### TLS Ciphers
+
+Zowe is always updating the ciphers used to follow industry best practice.
```
anaxceron, line 58: I think you need a blank lines here to create some line breaks.

docs/user-guide/tls-configuration.md
```diff
@@ -0,0 +1,85 @@
+# Customizing Native TLS
+
+Zowe's servers have built-in TLS support to enable HTTPS connections.
+This is the default, and an alternative to using AT-TLS which is [documented here](./at-tls-configuration).
+
+:::info Required roles: security administrator
```
anaxceron, line 6: Can we make "roles" singular?

anaxceron (commented):
Made some comments re: spacing and style, otherwise it looks good to me. But I think others should also review this before merging this pr.

docs/user-guide/configuring-at-tls-for-zowe-server.md
```diff
@@ -32,15 +32,15 @@ Follow these steps to configure Zowe to support AT-TLS:
...
-            attls: true
+  network:
+    # For inbound traffic rules:
+    server:
+      tls:
+        attls: true
+    # If outbound traffic rules will be configured:
+    client:
+      tls:
+        attls: true
```
1000TurquoisePogs, line 43: fixed this, it was a bug. it appeared to have more spaces then you'd want for yaml

docs/user-guide/configuring-at-tls-for-zowe-server.md
```diff
@@ -684,7 +684,3 @@ TTLSCipherParms                   CipherParms
 ```
 
 </details>
-
-## Additional Zowe feature configuration with AT-TLS
-
-The Zowe Application Framework also leverages AT-TLS. For more information, see [Using AT-TLS in the App Framework](../user-guide/mvd-configuration#using-at-tls-in-the-app-framework).
```
1000TurquoisePogs, line 690: removed this because not only does this header no longer exist, but that page referenced now just points back to this page regarding at-tls because this page is now the authority on at-tls.

docs/user-guide/tls-configuration.md
1000TurquoisePogs: Done thanks!

docs/user-guide/tls-configuration.md
1000TurquoisePogs: Done thanks!

docs/user-guide/configuring-at-tls-for-zowe-server.md
```diff
@@ -1,4 +1,4 @@
-# Configuring AT-TLS for Zowe Server
+# Enabling AT-TLS
```
pablocarle, line 1: If the title changes I think it should change the name of the file and the item in sidebars cc. @janan07

docs/user-guide/tls-configuration.md
```diff
@@ -0,0 +1,90 @@
...
+* TLSv1.3
+
+Zowe defaults to the following configuration:
+
+```yaml
+zowe:
+  network:
+    tls:
+      minTls: "TLSv1.2"
+      maxTls: "TLSv1.3"
```
pablocarle, line 56: Is this still exclusive with attls being enabled?
It's for a separate issue, but we think it should be allowed to customize these values even with at-tls enabled. Although there are other restrictions (such as ICSF hardware keys). We are not assuming all onboarded services will be AT-TLS, it should be still possible to use Java TLS for these scenarios even with AT-TLS enabled in Zowe.

docs/user-guide/configuring-at-tls-for-zowe-server.md
1000TurquoisePogs: I think changing filenames can break bookmarks and google indexing so in the past I have left filenames outdated just so users still end up on the right page.

docs/user-guide/tls-configuration.md
1000TurquoisePogs: The schema does have these as exclusive. Maybe the doc should make this clear?

It's a good point about a hybrid scenario. It's also difficult to make intuitive because people might think these settings would control AT-TLS when AT-TLS is enabled, but they're unrelated.

docs/user-guide/configuring-at-tls-for-zowe-server.md
pablocarle: Could be, we already changed this one once, since the original title was API ML only. I think the search is not working properly in stable version atm

docs/user-guide/tls-configuration.md
pablocarle: I think so if we are not planning on removing this exclusion, although I think we should.

docs/user-guide/configuring-at-tls-for-zowe-server.md
balhar-jakub: Just a note: if we want to change the names of the files, we should move the file and keep only the link to the new file on the original page.

balhar-jakub (approved):

PR #3724, opened 2024-06-21
add request buffer size doc

pablocarle:
Add new parameter to customize API ML websocket request buffer size


janan07 (approved):
Reviewed and edited by doc squad

PR #3760, opened 2024-07-10
Add internal api ml mapper config

janan07:
Describe your pull request here: Authenticating with client certificates was missing the procedure to enable the internal API mapper This PR addresses this omission.

List the file(s) included in this PR: authenticating-with-client-certificate.md

### DIFF
--- a/docs/user-guide/authenticating-with-client-certificates.md
+++ b/docs/user-guide/authenticating-with-client-certificates.md
@@ -23,7 +23,7 @@ When sending a request to a service with a client certificate, the Gateway perfo
 4. The Gateway then performs the login of the mapped user and provides valid authentication to the downstream service. 
 
 :::note Notes:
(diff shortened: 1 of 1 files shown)


docs/user-guide/authenticating-with-client-certificates.md
```diff
@@ -71,6 +71,17 @@ Register the client certificate with the user ID in your ESM. The following comm
 This metadata can be used for TLS client authentication.
 :::
 
+## Enable the internal API ML mapper
```
balhar-jakub, line 74: I don't think it belongs here, I believe the section should just note that the configuration information moved to the relevant Configuration article.

janan07: Ok. Thanks for taking a look. I'll just link to it in the client cert article.

PR #3852, opened 2024-09-03
Site PDF for Zowe v2.18

anaxceron:
Describe your pull request here: site PDF for Zowe v2.18

List the file(s) included in this PR: zowe-docs.pdf


zFernand0 (approved):
Thanks for updating the PDF :)

I'm a bit curious if we could compress these going forward.
I know it isn't much memory, but every PDF that we move into the `static` folder is about 40Mb. 
Using an online compressor seems to help (i.e. it cuts the size of each PDF in half), however, I don't know if this is the right thing to do here. 😅

PR #3887, opened 2024-09-23
V3 app framework updates  and misc.

1000TurquoisePogs:
I went through all but the tutorials & samples section of the app framework docs and updated them for v3.

In doing so I also removed a few lines that sounded like "this feature has been available since version 1.xx..." because by now those features are so old that in the v3 doc we dont need the trivia of when it was introduced, the changelogs have that.

While reviewing the overall server docs I noticed content from other squads docs that was missing and added what I know.


docs/extend/extend-desktop/mvd-server-config.md
```diff
@@ -1,13 +1,12 @@
 The Zowe's App Server and ZSS rely on many required or optional parameters to run, which includes setting up networking, deployment directories, plugin locations, and more. 
-These parameters can be specified in multiple ways: configuration files, CLI arguments, or environment variables.
+These parameters can be specified in two ways: configuration files, or environment variables.
```
anaxceron, line 2: Is this md file supposed to be in Zowe Docs? I ask b/c it's not included in the sidebar file, therefore it doesn't appear on the Zowe Docs site (although it _is_ in the Zowe Docs repository). If this is supposed to be included in the site, please add the file to the appropriate location in the `sidebar.js` file.

anaxceron (approved):

PR #3888, opened 2024-09-25
Mark caching service vsam mode deprecated

1000TurquoisePogs:
This PR updates documents referring to caching service's vsam storage mode.
This is deprecated in v3, simplifying installation, so I am trying to call less attention to this option so that users do not need to take an action on vsam as they used to.


balhar-jakub (approved):

janan07 (approved):
Reviewed by doc squad

PR #3892, opened 2024-09-27
Updated powerpoint and associated images for v3

1000TurquoisePogs:
This PR has the following changes:

* Updated architecture diagram terminology
    * ZLUX was used when App Server should be used
    * zssServer was used when ZSS should be used
    * Incorrect network arrows indicate ZSS typically connects directly to App-server. It does not. Only in the absence of the APIML gateway does that happen.

* Added missing v3 architecture information
    * ZAAS was missing from architecture diagram
    * ZAAS was missing from kubernetes diagram

* Removed architecture diagram files that had "copy" at the end of the name. I could find no references to these being used by anything.


janan07 (approved):
Reviewed by doc squad

PR #3906, opened 2024-10-01
Update v3_0_0.md for breaking changes content

1000TurquoisePogs:
The release notes page is not in sync with the contents of the https://github.com/zowe/docs-site/blob/docs-staging/docs/whats-new/breaking-changes-v3.md page.

This PR fixes this problem for the app framework, but makes the additional changes:

* Although the majority of the server notes were above the client notes, for some reason zss was not and was below the client notes. I have moved it up to be in the same section to avoid confusion.

* The ZSS note and call to action has been reworded to hopefully be more helpful

* Nobody should be calling the Zowe Server Install Wizard "ZEN" in the documentation. That is a code name not a program name.


anaxceron (approved):
Looks good, thank you @1000TurquoisePogs

PR #3915, opened 2024-10-03
Add missing security prerequisites for OIDC and document alternative ESM command for the x509 internal mapper

taban03:
* Add missing security prerequisites for OIDC
* Document alternative ESM command for the x509 internal mapper
* Modify note in the X.509 Client Certificate Authentication article for v3 to specify that the internal mapper API is the default method.
The changes are apply on both v2 and v3


versioned_docs/version-v2.18.x/user-guide/authenticating-with-client-certificates.md
```diff
@@ -52,7 +52,18 @@ Register the client certificate with the user ID in your ESM. The following comm
...
+  SETROPTS RACLIST(DIGTCERT, DIGTRING) REFRESH
+  ```
+  Alternatively, in case you are using the internal API ML mapper, you can use the following command:
+  
+  ```
+  RACDCERT ID(<userid>) MAP 
+  SDNFILTER('<subject's-distinguished-name-filter>')
+  WITHLABEL('<label>')
+  ```
+  **Note:**  The `RACDCERT MAP` command allows mapping a certificate to a user without adding the X.509 certificate to the ESM database. While this approach is more convenient, it provides less security compared to the `RACDCERT ADD` command, which stores the certificate directly in RACF, offering better control and protection.
```
pablocarle, line 66: I also think we should add the commands for TSS and ACF2 in one PR, and the note should be global to all three.

```suggestion
  ``` 
  RACDCERT ADD(<dataset>) ID(<userid>) WITHLABEL('<label>') TRUST
  SETROPTS RACLIST(DIGTCERT, DIGTRING) REFRESH
  ```

  Alternatively, in case you are using the internal API ML mapper, you can use the following command:
  
  ```
  RACDCERT ID(<userid>) MAP 
  SDNFILTER('<subject's-distinguished-name-filter>')
  WITHLABEL('<label>')
  ```

  **Note:**  The `RACDCERT MAP` command allows mapping a certificate to a user without adding the X.509 certificate to the ESM database. While this approach is more convenient, it could be considered less secure than adding the certificate to the ACID as it offers better control and protection.
```

pablocarle: We can keep this PR open until we finish working on the other 2 ESMs

versioned_docs/version-v2.18.x/user-guide/authenticating-with-client-certificates.md
```diff
@@ -52,19 +52,50 @@ Register the client certificate with the user ID in your ESM. The following comm
...
+  SDNFILTER('<subject's-distinguished-name-filter>')
+  WITHLABEL('<label>')
+  ```
+  **Note:**  The `RACDCERT MAP` command allows mapping a certificate to a user without adding the X.509 certificate to the ESM database. While this approach is more convenient, it could be considered less secure than adding the certificate to the ACID as it offers better control and protection.
 
   **Example command in ACF2:** 
 
   `INSERT <userid>.<certname> DSNAME('<dataset>') LABEL(<label>) TRUST`
 
-  **Example command in Top Secret:**
```
janan07, line 61: Should we state that using the internal API Mapper is the recommended approach?

balhar-jakub: We state it in some places, but it seems that not everywhere.

taban03: I will add the same note in this paragraph as well

janan07 (approved):
Reviewed and edited by doc squad

PR #3918, opened 2024-10-03
Fix webui and misc docs for v3

1000TurquoisePogs:
This PR builds upon #3917 to fix branch sync issues. While doing review I also fixed a reference for JWT and a header.


janan07 (approved):

zFernand0 (changes requested):
Requesting changes not because this PR is bad, but to prevent it from being merged for now.

I found a commit that technically contains all of the changes that need to be reverted which I missed on my PR.

Thanks for opening this PR as it helped me pinpoint which commit I missed.

janan07 (approved):

zFernand0 (approved):

anaxceron (approved):

PR #3929, opened 2024-10-08
Update systemrequirements-zos.md SDSF recommendations

1000TurquoisePogs:
While reading the system requirements I noticed we say SDSF is required, and then go on to say that it depends and is avoidable.

I rewrote this in a way I thought would explain better, by using a table showing which actions use SDSF and what alternatives can be used.

Further down the page, I removed a reference to node v16 since it isnt supported anymore.


versioned_docs/version-v2.16.x/user-guide/systemrequirements-zos.md
```diff
@@ -1,5 +1,5 @@
 ---
-keywords: [security permissions, system permissions, monacat]
```
1000TurquoisePogs, line 2: This change was from copy-pasting the v2.18 page.
Not sure what monacat is, seems to have been removed in newer versions of the doc.

balhar-jakub (approved):
That's a good change. Thanks.

PR #3935, opened 2024-10-08
Visualize "install" options as options under parent category

1000TurquoisePogs:
Today you can install zowe servers via many techniques and either on or off of z/OS.
All these options are currently visualized on the sidebar like this 
image

Since each one of these items is an option, you don't want to do more than one really. You want to choose one and only one.
To help make that obvious, I put them all under a parent item in the sidebar that just says "installing"
Then, I revised the titles to make clear what you're installing and how, so hopefully people will understand these are choices where you only choose one option.

Now it looks like this
image

### DIFF
--- a/sidebars.js
+++ b/sidebars.js
@@ -132,51 +132,57 @@ module.exports = {
         },
         {
           type: "category",
-          label: "Installing Zowe Server Install Wizard",
-          link: { type: "doc", id: "user-guide/install-zowe-server-install-wizard" },
+          label: "Installing",
           items: [
-            "user-guide/troubleshooting-zowe-server-install-wizard",
-          ]
-        },
-        {
-          type: "category",
-          label: "Installing Zowe via SMP/E",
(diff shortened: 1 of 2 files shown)


sidebars.js
```diff
@@ -133,51 +133,57 @@ module.exports = {
...
+                "user-guide/install-zowe-smpe-zosmf-workflow",
               ],
             },
-            "user-guide/install-zowe-smpe-zosmf-workflow",
-          ],
-        },
-        "user-guide/install-zowe-zos-convenience-build",
-        {
-          type: "category",
-          label: "Installing Zowe via a containerization build (PAX file)",
```
1000TurquoisePogs, line 173: Note this was a bug: the container build never was a pax file.

balhar-jakub:
This looks like a great improvement to me.

janan07 (approved):
This is good. It makes the installation options in the sidebar more legible.

Issue #3941, opened 2024-10-10
Document how to configure file associations for ZE data sets


t1m0thyj:
## Is your request for enhancement related to a problem? Please describe.

Currently we direct users to VS Code documentation about how to set up file associations, but in ZE V3 more detailed instructions are needed to define file associations that work for PDS members as well as sequential data sets.

## Describe the solution you'd like

We should suggest to define wildcards in the format `**/*COBOL*{,/*}` as suggested by @traeok in https://github.com/zowe/zowe-explorer-vscode/issues/3181#issuecomment-2393608608:

> Regarding the `files.associations` change in behavior: this is likely due to the change in path structure for PDS members. Before, we had inserted the member name into the file in parentheses (ex: `TEST.COBOL(MEMB1)`) vs. the new path structure `TEST.COBOL/MEMB1`. For v3, you'll need to adjust your associations so that they also support PDS members. Here are some example globs for reference:
> 
> ```
> **/*COBOL*{,/*} - matches /lpar.zosmf/TEST.COBOL.PS and /lpar.zosmf/TEST.COBOL.PDS/MEMBER
> **/*JCL*{,/*} - matches /lpar.zosmf/TEST.JCL.PS and /lpar.zosmf/TEST.JCL.PDS/EXAMPLE
> ```

## Related doc pages

https://docs.zowe.org/stable/user-guide/ze-usage-tips/#configure-the-detected-language-of-a-file-or-data-set

PR #3946, opened 2024-10-11
documenting file associations instructions for V3

anaxceron:
Describe your pull request here: Documenting new file associations instructions for V3 to address [Issue 3941](https://github.com/zowe/docs-site/issues/3941).

List the file(s) included in this PR: ze-usage-tips.md


t1m0thyj (approved):
Looks awesome, thanks @anaxceron!

docs/user-guide/ze-usage-tips.md
```diff
@@ -12,7 +12,17 @@ Zowe Explorer supports syntax highlighting for data sets and USS files. To enabl
...
+
+ To set file associations that work for sequential data sets as well as partitioned data set members, use wildcards with the language identifier in the format `**/*LANGUAGE*{,/*}`:
+
+```
+    "files.associations": {
+        "**/*COBOL*{,/*}": "cobol"
+    }
+```
+
+In the example above, Zowe Explorer uses wild cards to find matches of the configured language (`COBOL`) in the file paths of sequential data sets (for example, `/lpar.zosmf/TEST.COBOL.PS`) and PDS members (for example, `/lpar.zosmf/TEST.COBOL.PDS/MEMBER`).
```
anaxceron, line 25: @t1m0thyj @traeok: Should we say here that the instructions for V2 are different and either (1) link to them, or (2) write them out here?

t1m0thyj: In V2 the paths were structured differently, such that the data set name was always contained in the filename (or last segment of the file path). Paths for sequential data sets were formatted the same way (`/lpar.zosmf/TEST.COBOL.PS`) and for PDS members were different (`/lpar.zosmf/TEST.COBOL.PDS(MEMBER)`).

Because of this the wildcard pattern to match `COBOL` was significantly simpler - just `*COBOL*` would achieve the same effect in V2 as `**/*COBOL*{,/*}` in V3.

traeok: Although you can technically use a shorter glob for v2, could we recommend  the same one for both v2 and v3?

As far as I can tell, this glob will behave the same way as the shorter one. And this way, users can keep their file associations with this glob pattern and will be able to leverage them without issue if they jump from v2 -> v3 (e.g., once v2 is EOL)

traeok (approved):
Looks good to me, thanks Ana!

docs/user-guide/ze-usage-tips.md
t1m0thyj: I like this idea - although a simpler glob works in V2, the same glob will work in both V2 and V3, so we can probably avoid needless complexity by keeping the docs the same for both versions 😋

sbohnert (approved):
Looks good.

docs/user-guide/ze-usage-tips.md
anaxceron: ok, I won't add any further doc then re: v2 -- thank you

adam-wolfe (approved):
LGTM

PR #3977, opened 2024-11-04
Add missing details for APIML cookie config

1000TurquoisePogs:
When adding compatibility in app-server for APIML's customizable cookie name, I needed to look up how it worked.
I found the document does not fully describe "zowe.cookieIdentifier"
It instead referenced a "cookieIdentifier", without stating what that was or where to find it.
I confirmed APIML's code really is referring to "zowe.cookieIdentifier" https://github.com/search?q=repo%3Azowe%2Fapi-layer%20cookieidentifier&type=code which is present in both v2 and v3 https://github.com/zowe/zowe-install-packaging/blob/v3.x/staging/example-zowe.yaml#L317

I updated the document for each version and made clarifications to help explain it.


docs/user-guide/api-mediation/configuration-unique-cookie-name-for-multiple-zowe-instances.md
```diff
@@ -8,11 +8,14 @@ To prevent overwriting of the default cookie name in the case of multiple Zowe i
 
 Follow this procedure to configure a unique cookie name for the instances:
 
-1. Open the `zowe.yaml` configuration file.
-2. Find or add the property  `components.gateway.apiml.security.auth.uniqueCookie`, and set it to `true`. A unique cookie name is generated as `apimlAuthenticationToken.cookieIdentifier`.
+1. Open the Zowe YAML configuration file.
```
anaxceron, line 11: For the live site and previous versions:
- Why are we not saying `zowe.yaml` like we did before? 
- I think we should say how to open it, like: "Use a text editor to open the `zowe.yaml` configuration file."

docs/user-guide/api-mediation/configuration-unique-cookie-name-for-multiple-zowe-instances.md
```diff
@@ -8,11 +8,14 @@ To prevent overwriting of the default cookie name in the case of multiple Zowe i
...
+2. Find or add the property  `components.gateway.apiml.security.auth.uniqueCookie`, and set it to `true`.
```
anaxceron, line 12: If someone needs to add this property, will they know where to add it? If not, should we specify the location?

docs/user-guide/api-mediation/configuration-unique-cookie-name-for-multiple-zowe-instances.md
```diff
@@ -8,11 +8,14 @@ To prevent overwriting of the default cookie name in the case of multiple Zowe i
...
+3. Find the property `zowe.cookieIdentifier` and set it to any word or number that you wish to identify your Zowe instance.
+4. Restart Zowe.
+
+This will set a suffix for the API Gateway's cookie name based upon the value of `zowe.cookieIdentifier`.
```
anaxceron, line 16: For the live site and previous versions: Avoid future tense in documentation: "This sets a suffix ... "

docs/user-guide/api-mediation/configuration-unique-cookie-name-for-multiple-zowe-instances.md
```diff
@@ -8,11 +8,14 @@ To prevent overwriting of the default cookie name in the case of multiple Zowe i
...
 
     **Example:**  
-    If this parameter is set to `true`, and the cookieIdentifier is `1`, the name of the cookie transforms to `apimlAuthenticationToken.1`.  
-    If this property is not set to `true`, the cookie name remains `apimlAuthenticationToken` by default.
+    If `components.gateway.apiml.security.auth.uniqueCookie` parameter is set to `true` and `zowe.cookieIdentifier` is `1`, the cookie name will be `apimlAuthenticationToken.1`.
```
anaxceron, line 19: For the live site and previous versions:
- Please add a definite article: "If **the** `bbbb` parameter is set ... "
- Avoid future tense: " ... the cookie name is `blahblah`."

anaxceron (commented):
Thank you @1000TurquoisePogs. Made some comments re: clarity, grammar, style.

docs/user-guide/api-mediation/configuration-unique-cookie-name-for-multiple-zowe-instances.md
1000TurquoisePogs: Because zowe.yaml is incorrect since the YAML may not even be a file with a .yaml suffix

https://docs.zowe.org/stable/user-guide/configmgr-using#splitting-configuration-into-multiple-storage-types documents that YAML can be stored in PARMLIB
Such parmlibs wouldnt be named thing.yaml, because they aren't files in the traditional sense.
For lack of a better way of saying it, in docs I tend to say "Zowe YAML configuration file"

docs/user-guide/api-mediation/configuration-unique-cookie-name-for-multiple-zowe-instances.md
1000TurquoisePogs: I believe so, but think about this:
"components.gateway.apiml.security.auth" exists here https://github.com/zowe/zowe-install-packaging/blob/v3.x/staging/example-zowe.yaml#L502-L511
So we're saying to add a sub-attribute called "uniqueCookie"

In many places of the doc, we use this "this.that.thing" format for brevity and every time devs interact with users.
I believe it's well understood in general and by enough exposure by now, but the concept is covered in this note in the YAML reference:
https://docs.zowe.org/v2.18.x/appendix/zowe-yaml-configuration

The only thing I can think to do to make it more apparent is to somehow be able to reference that appendix in every place where we make such YAML references. But, I don't think anyone agreed upon a good way to do that if at all.

docs/user-guide/api-mediation/configuration-unique-cookie-name-for-multiple-zowe-instances.md
1000TurquoisePogs: Ok will fix.

docs/user-guide/api-mediation/configuration-unique-cookie-name-for-multiple-zowe-instances.md
anaxceron: @1000TurquoisePogs -- I think we can avoid linking to the appendix everywhere. We could link to the glossary since the definition includes a link to the appendix. I leave it to you, but my suggestion would be creating a link to the glossary term "Zowe configuration file":

1. Open the [Zowe YAML configuration file](https://docs.zowe.org/stable/appendix/zowe-glossary/#zowe-configuration-file)

janan07 (approved):
Reviewed by doc squad

PR #4029, opened 2024-12-05
Clarify jobname for app-server port

1000TurquoisePogs:
The table that shows which jobs to associate with which ports for Zowe server networking is missing detail on a quirk of NodeJS+zOS by which the jobname of the app-server that uses the stated port may not be what's intended, depending on configuration.

I've updated the table to show 2 jobnames for app-server, and then added a section directly below to clarify when and how these jobnames take affect, and what the defaults are.


1000TurquoisePogs:
Broken link checker reports
"* [520] [https://testssl.net/openssl-iana.mapping.html](https://testssl.net/openssl-iana.mapping.html) | Failed: Network error: Unknown status code"

This just means the site is broken right now. I assume it's some downtime and will return tomorrow.

ArooshLele (approved):
I approve of these updates and am signing off.

ArooshLele (approved):

ArooshLele (approved):

ArooshLele (approved):

ArooshLele (approved):

ArooshLele (approved):

PR #4047, opened 2025-01-03
Fix Netlify deploy for Dependabot PRs

t1m0thyj:

Updates build workflow to skips "Deploy to Netlify" step for Dependabot PRs. Since Dependabot doesn't have access to regular secrets, we can deploy w/o them same as we do for fork PRs.

The method for detecting that a PR is opened by Dependabot is based on [GitHub docs](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#fetching-metadata-about-a-pull-request).

List the file(s) included in this PR: build-docs.yml

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -58,7 +58,7 @@ jobs:
 
       - name: Deploy to Netlify
         id: deploy
-        if: ${{ github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository }}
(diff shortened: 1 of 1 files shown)


traeok (approved):
Changes make sense. I see that the PR instructions pipeline fails when only workflow files are updated - not sure if that's intended, but its not a big deal and it could be resolved in the future 😋

zFernand0 (approved):
Changes LGTM! 😋 
Thanks for sharing the GH docs link. 🙏

PR #4070, opened 2025-01-14
language refactor and question re: configuring AT-TLS for Zowe server

janan07:


docs/user-guide/configuring-at-tls-for-zowe-server.md
```diff
@@ -104,8 +112,9 @@ TTLSKeyringParms ZoweKeyring
   Keyring ZWEKRNG
 }
 ```
+<!--Can we provide a list of all parameters that need to be set in the previous example with the definitions of those parameters? -->
```
pj892031, line 115: I guess we shouldn't make a replacement of IBM documentation: https://www.ibm.com/docs/en/zos/3.1.0?topic=applications-tls-policy-statements

PR #4100, opened 2025-01-30
Fix ESM command to highlight what should be replaced

pj892031:
This PR adds missing symbols `<` and `>` to show what values should be updated before usage. It makes the documentation more readable.


janan07 (approved):
It appears that this PR doesn't affect descriptions in the file. Indicating variables for customization looks good.

pablocarle (approved):

pablocarle:
Will this PR be synched in master to apply to current version as well? Otherwise it will have to wait till 3.2 @janan07 @pj892031

PR #4110, opened 2025-01-31
fix: look for typedoc ending in .zip

traeok:

Recently, the typedoc was updated so that a `.zip.bundle` file is packaged alongside the other files in the Zowe release folder. However, this broke the grep check in `updateTypedoc.sh` that looks for the typedoc ZIP and it grabs the bundle instead, causing `unzip` to fail on line 25 of the script. This PR updates the check so it grabs the ZIP rather than the `.zip.bundle` file.

`scripts/updateTypedoc.sh`

### DIFF
--- a/scripts/updateTypedoc.sh
+++ b/scripts/updateTypedoc.sh
@@ -20,7 +20,7 @@ rm -rf typedoc
 rm -f zowe-nodejs-sdk-typedoc.zip
 
 # Download and extract new files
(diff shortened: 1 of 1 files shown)


anaxceron (approved):
thanks for the fix, @traeok!

t1m0thyj (approved):
LGTM, thanks @traeok!

Issue #3719, opened 2024-06-18
SWZELOAD is not documented


ericradz:
## Description
I noticed that in the Zowe install documentation, Step 5: Install the MVS data sets, indicates that after the execution of zwe install command, data sets SZWESAMP, SZWEAUTH and SZWEEXEC are created, but there is no mention of the SZWELOAD data set, which is also created in this step. There are other topics that mention only three of the four data sets too. Could you pls address this?

## Pages to Update
https://docs.zowe.org/stable/user-guide/install-zowe-zos-convenience-build/
https://docs.zowe.org/stable/user-guide/initialize-mvs-datasets/

There might be others as well.

## Screenshots
![image](https://github.com/zowe/docs-site/assets/36424317/986de797-ac57-4b8b-ae6e-4b60dfcdb129)

## Expected behavior
A description of the SZWELOAD data set to be added to the documentation in all of the appropriate topics.

1000TurquoisePogs:
I think the tables of datasets in the convenience build page should just be moved to an appendix and referenced wherever needed. Otherwise you will end up linking to the convenience build page in cases where the user isn't using the convenience build.

PR #4147, opened 2025-02-03
Create appendix for list of zowe's datasets

1000TurquoisePogs:
Resolves https://github.com/zowe/docs-site/issues/3719

This ticket collects information about zowe datasets and puts it in one appendix so that it can be referenced in a way that does not conflict with the user's chosen install type.

Before this PR, information about certain datasets was included (incompletely) in both SMPE and convenience build install pages.
While the tables were useful, referencing them from elsewhere could result in an issue where someone might be brought to a convenience build reference when using SMPE instead, or vice versa.

Putting the information into an appendix allows it to be referenced in a neutral way to both install types.

It also will be useful in the future for handling documentation about upgrades and downgrades as in https://github.com/zowe/community/issues/2305

This is a draft open to all feedback.
This would apply to both V2 and V3, but I havent made such updates without first gathering feedback.


1000TurquoisePogs:
Also:

Our code and doc call some datasets "custom datasets" and i find this bizarre terminology.
Maybe it's just me. I'd prefer saying they're "instance datasets", but wont make that change now unless there's consensus.

Martin-Zeithaml (approved):

ArooshLele (approved):

PR #4152, opened 2025-02-05
User/zlatnikp/apimlwfdoc

pavelzlatnik:
I have prepared doc changes related to new API ML configuration workflow which final version is part of Zowe V3 RC3.
Changes related to
https://github.com/zowe/zowe-install-packaging/pull/4144
https://github.com/zowe/zowe-install-packaging/pull/4154
https://github.com/zowe/zowe-install-packaging/pull/4161

When I was doing changes,
I have spot also few "bugs", forgotten updates from V2 dataset names, so I update it as well to reflect this doc is for V3, and dataset libraries should be with V3 naming convention, to reflect reality ( for example there is no sense to APF authorize dataset from V2 when user wants to run Zowe V3).

See list from this pull Request (like 18 files).


docs/user-guide/configure-apiml-zosmf-workflow.md
```diff
@@ -0,0 +1,113 @@
...
+
+:::info Required role: system programmer
+:::
+
+Ensure that you meet the following requirements before you start your Zowe configuration:
+
+- Install and configure z/OSMF
+- Install Zowe with an SMP/E build, PSWI, or a convenience build
+
+You can complete the following tasks with the **Configuration of API ML/Zowe 3.0** workflow:
```
janan07, line 16: Is this the correct name of the workflow? Configuration of API ML/Zowe 3.0

docs/user-guide/configure-apiml-zosmf-workflow.md
```diff
@@ -0,0 +1,113 @@
...
+- MVS Explorer
+- USS Explorer
+
+You can execute the Zowe configuration workflow either from a PSWI during deployment or later from a created software
+instance in z/OSMF. Alternatively, you can execute the configuration workflow z/OSMF during the workflow registration
+process.
+
+The configuration workflow described in this article is executed directly from a deployment of the Zowe PSWI.
+
+## Execute Configuration of Zowe 3.0 workflow from PSWI
```
janan07, line 41: Is this the correct name of the workflow or should this be API ML specific?

docs/user-guide/configure-apiml-zosmf-workflow.md
```diff
@@ -0,0 +1,113 @@
+# Configuring API ML with z/OSMF Workflows
+
+After you install Zowe, you can register and execute the z/OSMF workflows in the web interface to perform a range of
```
janan07, line 3: It seems that we should state that user's should use these workflows to configure API ML as a stand-alone component. Since no other configuration method exists to configure just API ML, I think we need to make sure that it is clear that this is the ONLY configuration method.

docs/user-guide/configure-apiml-zosmf-workflow.md
pavelzlatnik: right, change name to: Stand-alone Zowe API ML Configuration

docs/user-guide/configure-apiml-zosmf-workflow.md
pavelzlatnik: API ML specific would be probably better in here, so instead of "Configuration of Zowe 3.0"  to: Stand-alone Zowe API ML Configuration

docs/user-guide/configure-apiml-zosmf-workflow.md
pavelzlatnik: well, technically they can still use "normal" zowe configuration workflow, it is just way more complicated, if your goal is  API ML, so I would not say ONLY, but PREFERED

docs/user-guide/configure-apiml-zosmf-workflow.md
pavelzlatnik: updated

docs/user-guide/configure-apiml-zosmf-workflow.md
pavelzlatnik: updated

janan07 (approved):
Reviewed and refactored by doc squad

Issue #3984, opened 2024-11-05
Issue with docs.zowe.org/stable/user-guide/systemrequirements-zos


OnnoVdT:
## Description

`Node.js v18.x, or v20.x is required during installation.`
installation should be configuration. Node.js is not required to extract the pax or running SMP/E. These steps make up install, everything after is config.

## Pages to Update

 docs.zowe.org/stable/user-guide/systemrequirements-zos

## Expected behavior

Do not lump config in with install.

1000TurquoisePogs:
I will look into what's correct to say here. Yes, its not required for install, but what its required for is varied and less over time. zos squad is researching this.

1000TurquoisePogs:
I'm seeing occurrences of node used for the following purposes:

* Install of PAX, which uses "zwe install" which uses node to manipulate YAML
* Init using "zwe", zosmf workflows, or zowe server install wizard
    * Avoidable via use of JCL samples w/ Keyrings
* Startup when discovery enabled & zosmf used
* Containers
    * prestop - discovery registration cleanup is performed on stop
    * cleanup - cron job for doing registration cleanup periodically
* zwe support
* Most v2 zwe operations if zowe.useConfigmgr=false

1000TurquoisePogs:
node is needed for operations of this install page section https://docs.zowe.org/stable/user-guide/install-zowe-zos-convenience-build#step-5-install-the-mvs-data-sets

PR #4194, opened 2025-02-24
Update systemrequirements-zos.md  for node requirements

1000TurquoisePogs:
Resolves https://github.com/zowe/docs-site/issues/3984 by updating the docs.zowe.org/stable/user-guide/systemrequirements-zos page to state that zowe v3.1+ does not require node during installation.
Older zowe, and zowe v2, still have this requirement.


janan07 (approved):
Looks good

PR #4206, opened 2025-03-03
In the App-server Error Message Codes, the description of ZWED0003E is updated

ArooshLele:
On the App-server Error Message Codes page, I updated the description of the ZWED0003E.

appserver-error-codes.md


1000TurquoisePogs (approved):
I discussed this and the text here is accurate for what we want to tell users to do when they encounter this message.

PR #4276, opened 2025-04-02
reorder table and topics to group required config for API ML

janan07:
Describe your pull request here: reorder table and topics to group required config for API ML

List the file(s) included in this PR: configure-zos-system (v2.18), configuration-client-certificates, authenticating-with-client-certificates


versioned_docs/version-v2.18.x/user-guide/api-mediation/configuration-client-certificates.md
```diff
@@ -3,9 +3,9 @@
 :::info Roles: system programmer, system administrator, security administrator
 :::
 
-You can authenticate against API ML onboarded APIs. This functionality is disabled by default. Follow the steps in this article to enable authentication against API ML onboarded APIs. 
+In Zowe, you can authenticate against API ML onboarded APIs. This functionality is disabled by default. Follow the steps in this article to enable authentication against API ML onboarded APIs. 
 
-There are two methods to enable client certificate functionality: 
+There are two methods to enable x509 client certificate functionality:
```
balhar-jakub, line 8: Even though the certificates we use are x509 we also have x509 as the authentication scheme. 

For the authentication scheme you need neither setups. For the client certificate authentication in general you need either one of them. 

So maybe something like:
  There are two methods to enable client certificate functionality integrated with the SAF:

versioned_docs/version-v2.18.x/user-guide/api-mediation/configuration-client-certificates.md
```diff
@@ -3,9 +3,9 @@
 :::info Roles: system programmer, system administrator, security administrator
```
balhar-jakub, line 3: Generally be careful with addition of the x509, while it is true that the certificates we support are x509 we also use the term in another way which may confuse people.

janan07: The reason behind this addition is that there is a terminology discrepancy between references in Zowe Docs to just client certificates and references in ESZ that describe x.509 client certificates. It should be clear that the certificates are x509 in the Zowe doc if that is where we are sending ESZ customers.
x509 config section in ESZ

balhar-jakub: Ok, Then we should use X.509 as the exact name in the Zowe Docs as well.

balhar-jakub (approved):

Issue #3950, opened 2024-10-15
Explain how .zosattributes files can be used with Zowe CLI/Explorer


adam-wolfe:
## Is your request for enhancement related to a problem? Please describe.
The purpose/usage of .zosattributes files is not explained in existing documentation. The best explanation for how a .zosattributes file is supposed to work is in the webhelp here: https://docs.zowe.org/stable/web_help/index.html?p=zowe_zos-files_upload_dir-to-uss

...where it is described in a chunk of questionably formatted text at the very beginning of the (very long) command help for `zowe files upload dtu -h`.

## Describe the solution you'd like
In Zowe Docs, we should explain:
1. What is the purpose of a .zosattributes file?
2. How do I create a .zosattributes file?
3. Where do I put a .zosattributes file?
4. Do .zosattributes files work with Zowe Explorer?

PR #4337, opened 2025-04-28
Adding .zosattributes file doc

anaxceron:
Describe your pull request here: Addressing [Issue 3950](https://github.com/zowe/docs-site/issues/3950)

List the file(s) included in this PR: docs\user-guide\using-a-zos-attributes-file.md


docs/user-guide/using-a-zos-attributes-file.md
```diff
@@ -0,0 +1,68 @@
+# Using a z/OS attributes file
+
+Use a `.zosattributes` file to control how files are converted and tagged when you upload a local directory to a USS directory on the mainframe.
```
adam-wolfe, line 3: I think this also works for the download operation. Is that correct @zFernand0?

docs/user-guide/using-a-zos-attributes-file.md
```diff
@@ -0,0 +1,68 @@
...
+- `#`
+    - Denotes a comment when used at the start of a line. In the preceding example, comments are highlighted.
+
+- Each line can specify up to three positional attributes, or instructions:
+
+    - A pattern to match a set of files.
+        - Pattern-matching syntax follows the same rules as those that apply in `.gitignore` files. Negated patterns that begin with `!` are not supported.
+        - For syntax, see [Pattern Format]( https://git-scm.com/docs/gitignore#_pattern_format).
+    - A local-encoding to identify a file’s encoding on the local workstation.
+        - When `-` is specified for `local-encoding` **[should "local-encoding" be in back ticks?]**, files that match the pattern are not transferred.
```
adam-wolfe, line 39: I would not put local-encoding in back ticks.

docs/user-guide/using-a-zos-attributes-file.md
```diff
@@ -0,0 +1,68 @@
...
+
+```
+zowe zos-files upload dir-to-uss "local_dir" "/a/ibmuser/my_dir" --attributes my_global_attributes
+```
+
+`.zosattributes` files that are placed in nested directories are ignored.
+
+## Using a `.zosattributes` file with Zowe Explorer for VS Code
+
+In Zowe Explorer for Visual Studio Code, first upload a `.zosattributes` file to a USS directory on the mainframe and then upload files to that directory.
```
adam-wolfe, line 62: Shouldn't the `.zosattributes` file be in the folder on the desktop that is being uploaded?

docs/user-guide/using-a-zos-attributes-file.md
zFernand0: yeah, the `.zosattributes` file applies to both, upload and download 😋

docs/user-guide/using-a-zos-attributes-file.md
zFernand0: IIRC, it should work in both, but form a ZE perspective, I'm not sure having it just on the client dir is enough since one cannot upload an entire directory from ZE. 
I could be wrong though 😋

sbohnert (approved):

docs/user-guide/using-a-zos-attributes-file.md
adam-wolfe: Oh, right. Good point.

docs/user-guide/using-a-zos-attributes-file.md
```diff
@@ -0,0 +1,68 @@
...
+.* -
+// highlight-start
+# Upload jpg images in binary
+// highlight-end
+*.jpg binary binary
+// highlight-start
+# Convert CICS Node.js profiles to EBCDIC
+// highlight-end
+*.profile ISO8859-1 EBCDIC
+```
```
zFernand0, line 29: Wondering if we should just use a `shell` like highlighter for this block of code.
| Before | After |
| - | - |
| image | image |

Note: The `after` is just how GitHub displays _"```shell"_. Not sure if it will render the same way in the live site :yum:

zFernand0 (approved):
LGTM! 😋

zFernand0 (approved):

PR #4504, opened 2025-05-30
Documentation about memory assumption and limitation

pj892031:

### DIFF
--- a/docs/user-guide/address-storage-requirements.md
+++ b/docs/user-guide/address-storage-requirements.md
@@ -5,7 +5,7 @@
 
 Ensure that you have sufficient storage depending on the installation method. Review the storage requirements according to your installation method as presented in this article. 
 
-## Installing Zowe Server Rutntime
+## Installing Zowe Server Runtime
 
 Before installing Zowe, review the reference for [Zowe's server datasets](../appendix/server-datasets.md)
 
@@ -15,12 +15,73 @@ Additionally, when installing Zowe with SMP/E, review the [DASD storage requirem
 
 ## Memory requirements for API Mediation Layer
 
-Zowe API ML components have following memory requirements:
+Zowe API ML components as other Java application uses a various type of memories. Before you start limiting memory
+it is necessary to consider resource assumption. It depends also on you case and network traffic.
+
+The main type of memory is heap memory. The head memory is defined by initial size of memory and the maximum. Each
+service uses as defaul minimum 32MB and the maximum 512MB. When a service requires more memory it allocates more
+by increase of same size as initial memory. When the system has limited resources it is a good practice to se minimal
+same as the maximum. The memory is then allocated during the startup that ensure the memory cannot be exceeded by 
+different type of memory.
+
+Here you can see the expected heap memory requirements.
 
 Component name | Memory usage
 ---|---
 Gateway service | 256MB
+ZAAS | 256MB
 Discovery service | 256MB
 API Catalog | 512MB
 Caching service | 512MB
 
+All service uses also different type of memories, especially native, threads, buffers, and JIT (just in time compilation).
+In general there is an expectation that the total memory assumption should be more for 50% than heap size. Anyway as was 
+mentioned above it strictly depends on your use case (what data are transferred, how many users, what services are 
+onboarded, etc.). Therefor it is reasonable to verify your environment. It means run the API ML without limit, warn up 
+the system, and validate usage during the typical load. We also suggest to add some buffer to the maximum value.
+
+The buffers memory is as default limited to the same size as heap memory. It means it could allocate the same amount of
+memory in the case of big network traffic. It could be configured by argument `-XX:MaxDirectMemorySize=<value><k|K|m|M|g|G>`.
+There is possible to limit this memory for all service with the same configuration. It could be done by defining system
+environment `JAVA_OPTS`.
+
+As default threads responsible for processing incoming request are prepared just particular on start up. As default there
+are 20 threads and other are created on demand (as default up to 200). When the system has limited address space we 
+suggest to use thread pool with initiated all threads on the beginning. This behaviour could be done by configuring 
+properties `server.tomcat.threads.min-spare` (initial amount of threads) and `server.tomcat.threads.max` (the maximum
+amount of threads) to the same value.
+
+Example how to define memory by suggestions above:
+
+```yaml
+
+zowe:
+  environments:
+    SERVER_TOMCAT_THREADS_MIN_SPARE: 100
+    SERVER_TOMCAT_THREADS_MAX: 100
+    JAVA_OPTS: "-XX:MaxDirectMemorySize=64M"
+
+components:
+  gateway:
+    enabled: true
+    heap:
+      init: 256
+      max: 256
+  discovery:
+    enabled: true
+    heap:
+      init: 256
+      max: 256
+  api-catalog:
+    heap:
+      init: 512
+      max: 512
+  caching-service:
+    heap:
+      init: 512
+      max: 512
+  zaas:
+    heap:
+      init: 256
+      max: 256
+```
--- a/docs/user-guide/address-storage-requirements.md
+++ b/docs/user-guide/address-storage-requirements.md
@@ -3,7 +3,8 @@
 :::info Roles required: storage administrator, system programmer
 :::
 
-Ensure that you have sufficient storage depending on the installation method. Review the storage requirements according to your installation method as presented in this article. 
+Ensure that you have sufficient storage depending on the installation method. Review the storage requirements according 
+to your installation method as presented in this article. 
 
 ## Installing Zowe Server Runtime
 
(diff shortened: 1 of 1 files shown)


docs/user-guide/address-storage-requirements.md
```diff
@@ -15,12 +16,78 @@ Additionally, when installing Zowe with SMP/E, review the [DASD storage requirem
...
+Zowe API ML components as other Java application uses a various type of memories. Before you start limiting memory
+it is necessary to consider resource assumption. It depends also on your case and network traffic.
+
+The main type of memory is heap memory. The heap memory is defined by the initial size of the memory and the maximum. 
+Each service uses a default minimum of 32 MB and a maximum of 512 MB. When a service requires more memory it allocates 
+more by increasing the same size as the initial memory. When the system has limited resources it is a good practice to 
+set the minimum same as the maximum. The memory is then allocated during the startup which ensures the memory cannot 
+be exceeded by different kinds of memory.
+
+Here you can see the expected heap memory requirements.
```
pablocarle, line 28:
```suggestion
Zowe API ML components, like any other Java applications, uses various types of memory. Before starting to limit the system memory available to the Zowe API ML, it is necessary to consider resource consumption. 
The memory consumption depends on the use cases and network traffic.

The main type of memory used by Java applications is heap memory. The heap memory is defined by an initial size and a maximum. 
Each Zowe API ML service uses a default minimum of 32 MB and a maximum of 512 MB. When a service requires more memory, it allocates more memory in increments of the same size as the initial memory. When the system has limited resources, it is a good practice to set a fixed minimum and maximum, the memory is then allocated during startup, preventing the memory from exceeding the limit.

The following table shows the expected heap memory requirements for the core Zowe API ML services:
```

docs/user-guide/address-storage-requirements.md
```diff
@@ -15,12 +16,78 @@ Additionally, when installing Zowe with SMP/E, review the [DASD storage requirem
...
 ---|---
-Gateway service | 256MB
-Discovery service | 256MB
-API Catalog | 512MB
-Caching service | 512MB
+Gateway service | 512MB
+ZAAS | 128MB
+Discovery service | 128MB
+API Catalog | 128MB
+Caching service | 256MB
```
pablocarle, line 36:
```suggestion
Caching Service | 256MB
```

docs/user-guide/address-storage-requirements.md
```diff
@@ -15,12 +16,78 @@ Additionally, when installing Zowe with SMP/E, review the [DASD storage requirem
...
+
+Here you can see the expected heap memory requirements.
 
 Component name | Memory usage
 ---|---
-Gateway service | 256MB
-Discovery service | 256MB
-API Catalog | 512MB
-Caching service | 512MB
+Gateway service | 512MB
```
pablocarle, line 32:
```suggestion
Gateway Service | 512MB
```

docs/user-guide/address-storage-requirements.md
```diff
@@ -15,12 +16,78 @@ Additionally, when installing Zowe with SMP/E, review the [DASD storage requirem
...
+ZAAS | 128MB
+Discovery service | 128MB
```
pablocarle, line 34:
```suggestion
Discovery Service | 128MB
```

docs/user-guide/address-storage-requirements.md
```diff
@@ -15,12 +16,75 @@ Additionally, when installing Zowe with SMP/E, review the [DASD storage requirem
...
+`-XX:MaxDirectMemorySize=<value><k|K|m|M|g|G>`. Limit this memory could be set on the same value for all services. It 
+could be done by defining the system environment `JAVA_OPTS`.
+
+As default threads responsible for processing incoming requests are prepared particularly on start-up. As default, there
+are 20 threads and others are created on demand (as default up to 200). When the system has limited address space (see 
+usage of memlimit in the STC) we suggest using a threads pool with initiated all threads at the beginning. This behavior 
+could be done by configuring properties `server.tomcat.threads.min-spare` (initial amount of threads) and 
+`server.tomcat.threads.max` (the maximum amount of threads) to the same value.
+
+Example of how to define memory by the suggestions above:
```
pablocarle, line 53:
```suggestion
The following example shows all the above-mentioned parameters:
```

docs/user-guide/address-storage-requirements.md
```diff
@@ -15,12 +16,75 @@ Additionally, when installing Zowe with SMP/E, review the [DASD storage requirem
...
+The memory buffers are as default limited to the same size as heap memory. In the case of big network traffic, the JVM 
+could allocate up to the same amount of heap memory. Anyway, the memory limit could be configured by argument 
+`-XX:MaxDirectMemorySize=<value><k|K|m|M|g|G>`. Limit this memory could be set on the same value for all services. It 
+could be done by defining the system environment `JAVA_OPTS`.
+
+As default threads responsible for processing incoming requests are prepared particularly on start-up. As default, there
+are 20 threads and others are created on demand (as default up to 200). When the system has limited address space (see 
+usage of memlimit in the STC) we suggest using a threads pool with initiated all threads at the beginning. This behavior 
+could be done by configuring properties `server.tomcat.threads.min-spare` (initial amount of threads) and 
+`server.tomcat.threads.max` (the maximum amount of threads) to the same value.
```
pablocarle, line 51: I would split this in sections, like one for describing the the memory types and calculations needed and one for updating the settings.

Something like:

```suggestion
### JVM Memory

Different types of memories are used by JVM services: native, threads, direct memory buffers, and JIT (just-in-time compilation). 

The direct memory buffers are limited to the same size as heap memory. By default, in case of large network traffic, the JVM can allocate up to the same amount of heap memory.

Likewise, threads responsible for processing incoming requests are prepared during JVM start-up. By default, 20 threads are available to process incoming requests, with additional ones created on demand (up to 200 by default). When the system has a limited address space, it is recommended to use fixed-size thread pools, initialising all threads at the beginning.

As a general rule, there is an expectation that the total memory consumption should be 150% of the heap size. 

### Customizing memory limits in Zowe API Mediation Layer

Setting memory limits strictly depends on the use case (the kind and size of data that is typically transferred, how many users can be active simultaneously, how many and what kind of services are onboarded, etc.). 

If limits need to be set, it is recommended to verify your environment, to do this perform the following procedure:
1. Run the API ML without a memory limit
2. warm up the system
3. validate usage during a typical workload.

To set a direct memory buffer limit, set the following property:
* `zowe.environments.JAVA_OPTS`: `-XX:MaxDirectMemorySize=<value><k|K|m|M|g|G>`

Where `<value><k|K|m|M|g|G>` is the limit size and unit. i.e. `64M` is a 64 Megabyte limit.

To establish a fixed thread pool set the following properties to the same value:
* `zowe.environments.SERVER_TOMCAT_THREADS_MIN_SPARE`: initial amount of threads
* `zowe.environments.SERVER_TOMCAT_THREADS_MAX`: maximum amount of threads

**Note:** Updating these environment values will impact all java-based services running as part of the Zowe Server.

```

docs/user-guide/address-storage-requirements.md
```diff
@@ -15,12 +16,88 @@ Additionally, when installing Zowe with SMP/E, review the [DASD storage requirem
...
+As a general rule, there is an expectation that the total memory consumption should be 150% of the heap size. 
+
+### Customizing memory limits in Zowe API Mediation Layer
+
+Setting memory limits strictly depends on the use case (the kind and size of data that is typically transferred, how many users can be active simultaneously, how many and what kind of services are onboarded, etc.). 
+
+If limits need to be set, it is recommended to verify your environment, to do this perform the following procedure:
+1. Run the API ML without a memory limit
+2. warm up the system
+3. validate usage during a typical workload.
```
janan07, line 52: By what means is the usage validated? Is validated the correct word or is this "determine" the usage"? Please provide detail.

docs/user-guide/address-storage-requirements.md
```diff
@@ -15,12 +16,88 @@ Additionally, when installing Zowe with SMP/E, review the [DASD storage requirem
...
+
+To set a direct memory buffer limit, set the following property:
+* `zowe.environments.JAVA_OPTS`: `-XX:MaxDirectMemorySize=<value><k|K|m|M|g|G>`
+
+Where `<value><k|K|m|M|g|G>` is the limit size and unit. i.e. `64M` is a 64 Megabyte limit.
```
janan07, line 57: Is it evident to anyone performing this configuration as to what each of the unit types represents, or should we add a list that describes what each of these unit types is?

docs/user-guide/address-storage-requirements.md
pj892031: It means that you start the system and then check what is memory usage. By this value is determined the configuration.

docs/user-guide/address-storage-requirements.md
```diff
@@ -15,12 +16,88 @@ Additionally, when installing Zowe with SMP/E, review the [DASD storage requirem
...
+
+To establish a fixed thread pool set the following properties to the same value:
+* `zowe.environments.SERVER_TOMCAT_THREADS_MIN_SPARE`: initial amount of threads
+* `zowe.environments.SERVER_TOMCAT_THREADS_MAX`: maximum amount of threads
```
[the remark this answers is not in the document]
pj892031: We can add default values, for  `initial amount of threads` it is 20, and for `maximum amount of threads` 200, but I guess it is written above. Feel free to add if you miss these values there.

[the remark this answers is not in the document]
janan07: Is there a general description of a use case when someone would need to configure a higher number of threads?

docs/user-guide/address-storage-requirements.md
```diff
@@ -15,12 +16,88 @@ Additionally, when installing Zowe with SMP/E, review the [DASD storage requirem
...
+    heap:
+      init: 512
+      max: 512
+  zaas:
+    heap:
+      init: 256
+      max: 256
+```
+
+It is recommended to set `REGION=0M` in the STC. (default) and not set a `MEMLIMIT`. This will prevent issues caused by insufficient memmory.
```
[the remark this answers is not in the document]
pj892031: It is part of IBM documentation, ie.: https://www.ibm.com/docs/en/zos/3.1.0?topic=limit-specifying-region-size, https://www.ibm.com/docs/en/zos/3.1.0?topic=class-examples

We have just a few comments about, ie.:
https://docs.zowe.org/stable/user-guide/configure-uss/#address-space-region-size
https://docs.zowe.org/stable/user-guide/install-zowe-pswi-deployment/#resources-

I guess it is clear to customers. It is a JCL basics.

PR #4544, opened 2025-06-16
Simplify AT-TLS page

1000TurquoisePogs:
Some time ago @Joe-Winchester and I were looking at the state of the AT-TLS documentation and saw it could be improved now that Zowe overall covers AT-TLS - originally this page was just for APIML at a time when only APIML sufficiently handled AT-TLS.

Changes/Fixes:
* Ports 7551-7559 were referenced as Zowe's ports, but really only 7552-7558 currently are.
* ApimlKeyring was mentioned where elsewhere only ZoweKeyring was. Looks like a doc bug where not all occurrences were updated.
* For the z/OSMF rule, it was stated ZAAS (AZ) is the thing that needs the client rule, but elsewhere we see gateway (AG) rules for it - such rules now include both.
* The "southbound service" example I renamed as "extensions' server" example as in Zowe terminology, we can "install extensions" which may have their own servers/services. There is nuance here but I think "southbound service" is too generic so as to mislead people to think the example covers everything including Zowe core when it's really meant to explain how to add non-zowe servers into the AT-TLS environment.  If you have better ways to phrase this go ahead and make edits.
* ApimlClientRule is now ZoweClientRule - there was no rule that allowed for example, app-server to register to discovery

Edits welcome but I am hoping for a particular outcome - I would like to separate the bare minimum needed for all of Zowe to work under AT-TLS from different scenarios like "make only apiml work" or "add extensions". ATTLS is wildly complicated and the more we detail in a single page the harder it becomes for people to understand what the bare minimum is to get ATTLS working for Zowe. Perhaps even, move all the extra examples to a separate page? ATTLS can get as advanced as you need it to be but we can save that for advanced pages.


pablocarle:
> For the z/OSMF rule, it was stated ZAAS (AZ) is the thing that needs the client rule, but elsewhere we see gateway (AG) rules for it.

Actually, it should be both, given that ZAAS uses it as authentication provider, but Gateway may route to it

pablocarle:
"Southbound service" is a term we still use in API ML routing terminology, as a service that sits behind the gateway's routing but I suppose any generic term is good as long as it's more understandable

pablocarle:
@1000TurquoisePogs does this need to be backported to v2 doc as well? The articles are pretty much the same except for the ZAAS inclusion

docs/user-guide/configuring-at-tls-for-zowe-server.md
```diff
@@ -200,8 +204,8 @@ TTLSRule ApimlClientRule
   LocalAddr All
   LocalPortRange 1024-65535
   RemoteAddr All
-  RemotePortRange 7551-7555 # Range covers API ML services (gateway, zaas, discovery, api catalog, caching service)
-  Jobname ZWE1A* # Generate according to zowe.job.prefix in zowe.yaml
+  RemotePortRange 7552-7558 # Range covers API ML, app-server, and zss services
+  Jobname ZWE1AG # Generate according to zowe.job.prefix in zowe.yaml
```
pablocarle, line 208: This should be generic to apply to all Zowe core services

1000TurquoisePogs:
> @1000TurquoisePogs does this need to be backported to v2 doc as well? The articles are pretty much the same except for the ZAAS inclusion

Yes - I guess this can fan-out to all zowe releases except in v2 we need to change for
* lack of ZAAS
* no jobs-api/files-api support

1000TurquoisePogs:
> "Southbound service" is a term we still use in API ML routing terminology, as a service that sits behind the gateway's routing but I suppose any generic term is good as long as it's more understandable

Okay. Let's see if we can accommodate both terminology. Such as

> ...Southbound services, such as Zowe extensions or other servers outside of Zowe core...

1000TurquoisePogs:
@pablocarle I captured both gateway & zaas by rules using jobname `ZWE1A*`. It's simple but also captures other APIML such as api-catalog. Do you think that's bad for an example? The descriptions provide the full explanation.

docs/user-guide/configuring-at-tls-for-zowe-server.md
1000TurquoisePogs: Do you mean you think the rule should be for all zowe servers to talk to each other?

Technically a rule like `jobname zwe1*` and `remoteportrange 7552-7558` does work, and covers more cases (like only zss + app-server running) but is also broader than necessary for apiml use (servers should only talk to gateway & discovery)?

ArooshLele:
FYI: @ArooshLele  @gauravs-20  @Gautham-coder  @DivergentEuropeans

Issue #4416, opened 2025-05-09
Issue with docs.zowe.org/stable/user-guide/cli-using-understand-profiles-configs


colinpaicemq:
## Description

Its says

A project configuration file resides in a directory of your choice. It contains project user profiles and project team profiles.

I cannot find where to specify this directory of my choice.

PR #4981, opened 2026-04-13
adding commands to specify project config location

anaxceron:
Describe your pull request here: Addressing [Issue 4416](https://github.com/zowe/docs-site/issues/4416)

List the file(s) included in this PR: cli-using-understand-profiles-configs.md

### DIFF
--- a/docs/user-guide/cli-using-understand-profiles-configs.md
+++ b/docs/user-guide/cli-using-understand-profiles-configs.md
@@ -16,19 +16,22 @@ If configuration files were used in the example above, the user would have neede
 
 `zowe zos-files list data-set "SYS1.PARMLIB*"`
 
-## Learning the terminology
+## Zowe client side terminology
 
 As of Zowe V2, Zowe CLI relies on the profiles stored in configuration files to obtain connection information.
 
 Both *user* and *team* profiles are stored in configuration files, and these configuration files can either be *project* configuration files or *global* configuration files. It is helpful to understand how these differ.
 
 - A **user configuration file** stores *user profiles* and is used for one person who needs their own unique properties to run commands.
+    - Created with the command `zowe config init --user-config`.
 
 - A **team configuration file** stores *team profiles* and is used by a group of people who need the same properties to run commands.
+    - Created with the command `zowe config init`.
 
 - A **global configuration file** resides in the `ZOWE_CLI_HOME` directory (YourUserHomeDirectory/.zowe, by default). It contains global *user profiles* and global *team profiles*.
-
+    - Created with the commands `zowe config init --global-config` or `zowe config init --user-config --global-config`.
 - A **project configuration file** resides in a directory of your choice. It contains project *user profiles* and project *team profiles*.
+    - Created with the command `zowe config init`. There can be multiple project configuration files. To define the directory for a configuration file, issue the `zowe config init` command from the directory you want the project configuration applied. 
 
 All configuration files are saved in `.json` format.
 
@@ -38,7 +41,7 @@ There may be instances where a user has all four types of files in their system,
 
 This can mean working with files that have conflicting configurations. One file can specify that a certain profile property has a value of `ABC`, while another file uses `XYZ` as a value.
 
-When the same properties have different values across multiple configuration files, Zowe CLI follows a two-step check to determine which configurations apply:
+When the same properties have different values across multiple configuration files, Zowe CLI looks for the narrowest scope with the most specific application by following a two-step check:
 
 1. Does the configuration file have a *more narrow* or a *more broad* scope?
 2. Is the configuration file *more specific* or *less specific*?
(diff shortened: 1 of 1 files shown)


docs/user-guide/cli-using-understand-profiles-configs.md
```diff
@@ -16,19 +16,22 @@ If configuration files were used in the example above, the user would have neede
...
+    - Created with the command `zowe config init --user-config`.
 
 - A **team configuration file** stores *team profiles* and is used by a group of people who need the same properties to run commands.
+    - Created with the command `zowe config init`.
 
 - A **global configuration file** resides in the `ZOWE_CLI_HOME` directory (YourUserHomeDirectory/.zowe, by default). It contains global *user profiles* and global *team profiles*.
-
+    - Created with the commands `zowe config init --global-config` or `zowe config init --user-config --global-config`.
 - A **project configuration file** resides in a directory of your choice. It contains project *user profiles* and project *team profiles*.
+    - Created with the command `zowe config init`. There can be multiple project configuration files. To define the directory for a configuration file, issue the `zowe config init` command from the directory you want the project configuration applied.
```
zFernand0: I understand we are trying to respond to this query:

> "I cannot find where to specify this directory of my choice."

However, I do find this a bit strange. What does it mean to "specify the directory" for a configuration file? It feels a little bit backwards since the configuration file is inside the directory. 

---

NPM has a similar configuration file. 
_Source_: https://docs.npmjs.com/cli/v8/configuring-npm/npmrc#per-project-config-file

What do we think about the following suggestion:


```suggestion
    - Created with the command `zowe config init`. There can be multiple project configuration files. When working locally in a project, a configuration file created with the `zowe config init` command will determine the connection details used in that project.
```

docs/user-guide/cli-using-understand-profiles-configs.md
```diff
@@ -38,7 +41,7 @@ There may be instances where a user has all four types of files in their system,
 
 This can mean working with files that have conflicting configurations. One file can specify that a certain profile property has a value of `ABC`, while another file uses `XYZ` as a value.
 
-When the same properties have different values across multiple configuration files, Zowe CLI follows a two-step check to determine which configurations apply:
+When the same properties have different values across multiple configuration files, Zowe CLI looks for the narrowest scope with the most specific application by following a two-step check:
```
zFernand0, line 44: > looks for the narrowest scope with the most specific application

It might be just me, but the above feels a bit excessive. 

Will it help to reverse the order of this sentence?
```suggestion
Zowe CLI follows a two-step check when the same properties have different values across multiple configuration files.
```

If I'm the only one confused by this, please disregard 🙏 
or: Please disregard if I'm the only one confused by this. 😅

zFernand0 (approved):
Changes LGTM! 😋 
---

I left some comments because I was just confused by the original request in the first place. That said, I do approve of these changes since the commands that generate each config are certainly valuable 🙏

docs/user-guide/cli-using-understand-profiles-configs.md
gejohnston: I am going to take a different swing at replacing the paragraph "- Created with the command ...". We might improve my wording below, but my intent is to convey that users determine and create their own directories for their project-specific work. Zowe simply lets you create different Zowe config files in each of those project directories.

- When your work on different projects is separated into different directories, you can create different Zowe configuration files in each such project directory by running the `zowe config init` command in each directory. Such Zowe project configuration files only affect your actions while you are within that particular directory.

gejohnston (commented):
I think we may want to revisit the currenly unchanged section titled "Zowe client side terminology". 

For a global config, we state that you can create both team profiles and user profiles (which is correct). Config files in a project directory have the same combination of two items (team and user). However, commands that will create project-level configuration files are categorized in the doc as 3 separate items: "user config", "team config", and "project config". It feels like this would be confusing to readers.

docs/user-guide/cli-using-understand-profiles-configs.md
anaxceron: My concern w/ the original wording is that it didn't clarify that the two-step check was looking for narrower/more specific configs. The original wording simply said, "Zowe CLI follows a two-step check to determine which configurations apply:". To me, it's important to clarify what exactly Zowe CLI is looking for when doing this check.

docs/user-guide/cli-using-understand-profiles-configs.md
anaxceron: @zFernand0 @gejohnston: From what I understand, the point of the issue is that we need to explain the steps a user needs to follow to designate a config file as a "project" config file (versus a global file). 

Edited the text to create yet another draft for this sub-bullet, please lmk your thoughts:

- Created with the command `zowe config init`. There can be multiple project configuration files. To designate a directory for a configuration file, issue the `zowe config init` command from the directory you want the project configuration applied. Project configuration files affect actions performed only while you are within that file's designated directory.

adam-wolfe (approved):
LGTM, thanks @anaxceron

gejohnston (approved):

PR #5000, opened 2026-04-29
secure certificate doc

anaxceron:
Describe your pull request here: Documenting [PR 2688](https://github.com/zowe/zowe-cli/pull/2688)

List the file(s) included in this PR: docs\user-guide\cli-authentication-methods.md

### DIFF
--- a/docs/user-guide/cli-authentication-methods.md
+++ b/docs/user-guide/cli-authentication-methods.md
@@ -293,9 +293,63 @@ See the [Integrating with API Mediation Layer](https://docs.zowe.org/zowe-docs-v
 
 Certificates are a long lasting type of authentication, rather than a password or token that can expire in hours, days, or months. A certificate is authenticated by matching a public and private key.
 
-To use a client certificate for authentication:
+Zowe CLI supports certificates stored in your operating system keystore and certificates stored in a local file.
 
-1. Specify the path to the certificate file in the relevant profile :
+:::tip
+
+If you have properties for multiple authentication methods stored in the same team configuration profile, set the `authOrder` property to your preferred [order of precedence](#order-of-precedence). Place `cert-pem` first in the list if you want Zowe CLI to check for that authentication method first.
+
+:::
+
+### Using certificates from keystore
+
+:::note
+
+This option is not supported for Linux operating systems.
+
+:::
+
+To use a certificate from keystore for authentication:
+
+1. If your certificate is not already imported into the keystore, import the certificate into your operating system certificate manager.
+
+2. Specify the name of the certificate in the relevant profile:
+
+   ```
+   zowe config set profiles.base.properties.certAccount <certName> 
+   ```
+
+   - `<certName>`
+   
+      Specifies the name of the certificate in your keystore. Matches the Common Name (CN) of the certificate.
+      
+    If you are using a **global** team configuration file (located in your home directory), add the `--global-config` option to the end of the command.
+
+    The configuration file profile is updated with the specified certificate Common Name:
+
+    ```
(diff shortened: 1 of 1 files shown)


JillieBeanSim (approved):
LGTM! thanks @anaxceron for the updates

docs/user-guide/cli-authentication-methods.md
```diff
@@ -293,9 +293,63 @@ See the [Integrating with API Mediation Layer](https://docs.zowe.org/zowe-docs-v
 
 Certificates are a long lasting type of authentication, rather than a password or token that can expire in hours, days, or months. A certificate is authenticated by matching a public and private key.
 
-To use a client certificate for authentication:
+Zowe CLI supports certificates stored in your operating system keystore and certificates stored in a local file.
```
t1m0thyj: Question for @adam-wolfe: Should we mention it's a prerequisite that the client certificate has been:
1. Created on z/OS with RACF or other ESM
2. Configured for the service being connected to
3. Downloaded to their local machine

If so, I don't think we should go into detail because exact steps will vary greatly (same as the process of importing into local cert store). But perhaps we could provide a link to z/OSMF doc as an example: https://www.ibm.com/docs/en/zos/3.2.0?topic=configurations-configuring-zosmf-server-certificate-key-ring

Or perhaps this addition would be unnecessary and out of scope, if you think its ok to assume that folks know how to obtain a client certificate as the article previously did?

t1m0thyj (approved):
LGTM, thanks @anaxceron! Left an optional comment/question 😋

docs/user-guide/cli-authentication-methods.md
```diff
@@ -293,9 +293,63 @@ See the [Integrating with API Mediation Layer](https://docs.zowe.org/zowe-docs-v
...
 
-1. Specify the path to the certificate file in the relevant profile :
+:::tip
+
+If you have properties for multiple authentication methods stored in the same team configuration profile, set the `authOrder` property to your preferred [order of precedence](#order-of-precedence). Place `cert-pem` first in the list if you want Zowe CLI to check for that authentication method first.
+
+:::
+
+### Using certificates from keystore
```
adam-wolfe: Is keystore an understood term in this context? It might be good to explain that when we are talking about a keystore, we are talking about operating system certificate storage.

docs/user-guide/cli-authentication-methods.md
adam-wolfe: It might be helpful to explain where these certificates come from. i.e., so that it's clear we are talking about certificates generated on the mainframe that a user can use to authenticate to the ESM and access a given service (like the z/OSMF REST APIs).

I wouldn't go into the details of generating certificates or importing them.

docs/user-guide/cli-authentication-methods.md
t1m0thyj: Would it be clearer to say "Using certificates from OS keychain"?

docs/user-guide/cli-authentication-methods.md
t1m0thyj: @anaxceron Something like this would be a good concise way to say it:

> Zowe CLI supports certificates stored in your operating system keystore and certificates stored in a local file. These are client certificates that must be generated on z/OS and allow users to authenticate to a given service such as z/OSMF.

docs/user-guide/cli-authentication-methods.md
adam-wolfe: I think so.

adam-wolfe (approved):

PR #5011, opened 2026-05-06
Add details for binding internal and external listeners to one stack and one port

janan07:
Describe your pull request here: Add details for binding internal and external listeners to one stack and one port

List the file(s) included in this PR: address-network-requirements.md

### DIFF
--- a/docs/user-guide/address-network-requirements.md
+++ b/docs/user-guide/address-network-requirements.md
@@ -17,7 +17,7 @@ Each Jobname has a default prefix of ZWE1, but that can be customized via the `z
 :::note
 **Internal and External Port Isolation**
 
-Zowe v3 provides isolation between internal and external listeners, enabling you to implement firewall rules that prevent external users from accessing sensitive management endpoints. For details, see [Binding to Specific or Multiple Addresses](#binding-to-specific-or-multiple-addresses) later in this article.
+Zowe v3 supports the use of separate internal and external listeners to isolate service-to-service communication from external traffic. For information on how to configure these listeners and their associated ports, see [Interface Binding and Traffic Isolation](#interface-binding-and-traffic-isolation) later in this article.
 :::
 
 ### Single-service deployment
--- a/docs/user-guide/address-network-requirements.md
+++ b/docs/user-guide/address-network-requirements.md
@@ -97,9 +97,9 @@ Defines the address used for internal listeners (service-to-service communicatio
 In environments with multiple TCP/IP stacks, you can bind Zowe to specific addresses across those stacks by providing a comma-separated list in the `zowe.network.server.listenAddresses` property. Zowe startup scripts automatically parse this list to ensure the services bind correctly to every specified interface.
 :::
 
-### TCPIP Port Reservations
+### TCP/IP Port Reservations
 
-It is also recommended to use TCPIP port assignment statements to restrict the IP and ports of each server by their jobnames. The jobnames of each Zowe component is derived from the property `zowe.job.prefix` and <component-suffix> as shown in the table prior. For details, see the topic _PROFILE.TCPIP port assignments_ in the IBM documentation.
+The best practice is to use TCP/IP port assignment statements to restrict the IP and ports of each server by their jobnames. The jobnames of each Zowe component is derived from the property `zowe.job.prefix` and `<component-suffix>`. For details, see the topic _PROFILE.TCPIP port assignments_ in the IBM documentation.
 
 When `zowe.job.prefix` is `ZWE1`, an example of port reservations with a fixed IP of `10.11.12.13` could be as follows with single-service API ML deployment:
 
@@ -112,7 +112,7 @@ When `zowe.job.prefix` is `ZWE1`, an example of port reservations with a fixed I
 ```
 
 :::note
-This TCP IP setting is valid for the Zowe Server started with `JOBNAME=ZWE1SV` option, for example `S ZWESLSTC,JOBNAME=ZWE1SV`.
+This TCP/IP setting is valid for the Zowe Server started with `JOBNAME=ZWE1SV` option, for example `S ZWESLSTC,JOBNAME=ZWE1SV`.
 :::
 
 With multi-service deployment, it could be as follows:
@@ -128,4 +128,4 @@ With multi-service deployment, it could be as follows:
    7558 TCP ZWE1AZ BIND 10.11.12.13 ; Zowe ZAAS
 ```
 
-:::note This TCP IP setting is valid for the Zowe Server started with `JOBNAME=ZWE1SV` option, for example `S ZWESLSTC,JOBNAME=ZWE1SV`. :::
+:::note This TCP/IP setting is valid for the Zowe Server started with `JOBNAME=ZWE1SV` option, for example `S ZWESLSTC,JOBNAME=ZWE1SV`. :::
--- a/docs/user-guide/address-network-requirements.md
+++ b/docs/user-guide/address-network-requirements.md
@@ -74,11 +74,11 @@ If you need to restrict Zowe to specific TCP/IP stacks or separate internal serv
 * **zowe.network.server.listenAddresses**  
 Defines the specific IP address or a comma-separated list of addresses for the server to bind to.
 
-* **zowe.network.external.address**  
-Defines the address used for external listeners (end-user traffic).
+* **zowe.externalPort**  
+Specifies the primary port number used for external user access to Zowe services. This is a single integer that defines the entry point for end-user traffic, typically routed through the API Gateway.
 
-* **components.apiml.internal.discovery.address** (used in single-service deployment)   
-Defines the address used for internal listeners for service-to-service communication. In a multi-service deployment, use `components.gateway.server.internal.address`. 
+* **zowe.externalDomains** (used in single-service deployment)   
+Specifies authorized domain names or hostnames that are valid for use with the Zowe instance. This property helps define the network boundaries and is used for security validations, such as ensuring that the Zowe instance only accepts requests directed to these specific domains.
 
 
 :::note Internal and External Listener Isolation
@@ -93,15 +93,14 @@ zowe:
     server:
       # Binds to both local and external interfaces
       listenAddresses: "127.0.0.1, 10.11.12.13"
-    external:
-      address: "10.11.12.13"
+  
+  # The main external port for end-user traffic
+  externalPort: 7554
 
-components:
-  apiml:
-    internal:
-      discovery:
-        # Defines the internal listener for the Discovery service
-        address: "127.0.0.1"
+  # List of valid domains for the Zowe instance
+  externalDomains: 
+    - "zowe.example.com"
+    - "api.internal.org"
 ```
 
 :::note Binding to Multiple TCP/IP Stacks 
@@ -113,7 +112,11 @@ In environments with multiple TCP/IP stacks, you can bind Zowe to specific addre
 
 The best practice is to use TCP/IP port assignment statements to restrict the IP and ports of each server by their jobnames. The jobnames of each Zowe component is derived from the property `zowe.job.prefix` and `component-suffix`. For details, see the topic _PROFILE.TCPIP port assignments_ in the IBM documentation.
 
-When `zowe.job.prefix` is set to `ZWE1` and the IP address is `10.11.12.13`, the port reservations for a **single-service API ML deployment** would appear as follows:
+When `zowe.job.prefix` is set to `ZWE1`, the external IP is `10.11.12.13`. The port reservations should be configured to reflect traffic isolation.
+
+**Single-Service Deployment**
+
+In a single-service deployment, Zowe components run under a single started task jobname (default `ZWE1SV`).
 
 **Example:**
 
@@ -125,11 +128,9 @@ When `zowe.job.prefix` is set to `ZWE1` and the IP address is `10.11.12.13`, the
    7557 TCP ZWE1SZ BIND 10.11.12.13 ; Zowe ZSS
 ```
 
-:::note
-This TCP/IP setting is valid for the Zowe Server started with `JOBNAME=ZWE1SV` option, for example `S ZWESLSTC,JOBNAME=ZWE1SV`.
-:::
+**Multi-Service Deployment**
 
-For a **multi-service deployment**, the port reservations would be configured as shown in this example:
+In a multi-service deployment, each component runs in its own address space with a unique jobname suffix.
 
 **Example:**
 
@@ -144,6 +145,9 @@ For a **multi-service deployment**, the port reservations would be configured as
    7558 TCP ZWE1AZ BIND 10.11.12.13 ; Zowe ZAAS
 ```
 
-This TCP/IP setting is valid for the Zowe Server started with `JOBNAME=ZWE1SV` option, for example `S ZWESLSTC,JOBNAME=ZWE1SV`. 
+:::note
+This TCP/IP setting is valid for the Zowe Server started with `JOBNAME=` parameter to match the prefix. For example `S ZWESLSTC,JOBNAME=ZWE1SV`.
+:::
 
-z
+
+
--- a/docs/user-guide/address-network-requirements.md
+++ b/docs/user-guide/address-network-requirements.md
@@ -127,7 +127,7 @@ When `zowe.job.prefix` is set to `ZWE1`, the external IP is `10.11.12.13`. The p
 
 **Single-Service Deployment**
 
-In a single-service deployment, API ML sub-components run under a single started task jobname (default `ZWE1AG`).
+In a single-service deployment, API ML components run under a single started task jobname (default `ZWE1AG`).
 
 **Example:**
 
@@ -141,7 +141,7 @@ In a single-service deployment, API ML sub-components run under a single started
 
 **Multi-Service Deployment**
 
-In a multi-service deployment, each component runs in its own address space with a unique jobname suffix.
+In a multi-service deployment, each API ML component runs in its own address space with a unique jobname suffix.
 
 **Example:**
 
--- a/docs/user-guide/address-network-requirements.md
+++ b/docs/user-guide/address-network-requirements.md
@@ -69,12 +69,13 @@ Configure how Zowe identifies and binds to network interfaces on your system. Th
 
 By default, Zowe binds to the wildcard address `0.0.0.0`, making Zowe services available on all network interfaces. This configuration is recommended for environments where strict interface isolation is not required.
 
-Zowe V3 utilizes a dual-listener architecture to separate internal traffic from user traffic. If you need to restrict Zowe to specific TCP/IP stacks or separate internal service traffic, use the following properties in zowe.yaml:
+Zowe V3 uses two listeners to separate internal traffic from external traffic. If you need to restrict Zowe to specific TCP/IP stacks or separate internal service traffic, use the following properties in zowe.yaml:
 
 * **Internal Port**  
-Used exclusively for Zowe services to register themselves with the API Mediation Layer and for service-to-service discovery. This traffic should be restricted to the local or internal network.
+Used for services to register themselves with the API Mediation Layer. This traffic is typically restricted to the internal network.
+
 * **External Port**  
-The primary gateway for all external user access (Web UI, CLI, VS Code) and onboarded services. This is the port used for opening firewalls and defining external access rules.
+The primary access point for all external user access (Web UI, CLI, VS Code) and onboarded services. This is the port used for opening firewalls and defining external access rules.
 
 To manage this isolation, use the following properties in zowe.yaml:
 
@@ -126,7 +127,7 @@ When `zowe.job.prefix` is set to `ZWE1`, the external IP is `10.11.12.13`. The p
 
 **Single-Service Deployment**
 
-In a single-service deployment, Zowe components run under a single started task jobname (default `ZWE1SV`).
+In a single-service deployment, API ML sub-components run under a single started task jobname (default `ZWE1AG`).
 
 **Example:**
 
--- a/docs/user-guide/address-network-requirements.md
+++ b/docs/user-guide/address-network-requirements.md
@@ -17,7 +17,7 @@ Each Jobname has a default prefix of ZWE1, but that can be customized via the `z
 :::note
 **Internal and External Port Isolation**
 
-Zowe v3 supports the use of separate internal and external listeners to isolate communication for registration of the services to API ML from external traffic. For information on how to configure these listeners and their associated ports, see [Interface Binding and Traffic Isolation](#interface-binding-and-traffic-isolation) later in this article.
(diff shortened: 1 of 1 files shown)


docs/user-guide/address-network-requirements.md
```diff
@@ -57,11 +63,60 @@ The Caching Service will use these additional ports if enabled (`components.cach
...
+
+If the default is not desired, you can use the following properties in zowe.yaml to restrict which network interfaces Zowe services use:
+
+* **zowe.network.server.listenAddresses**  
+Defines the specific IP address or addresses for the server to bind to. To bind to multiple interfaces or specific TCP/IP stacks, provide a comma-separated list of IP addresses.
+
+* **zowe.network.external.address**  
+Defines the address used for external listeners (end-user traffic).
+
+* **zowe.network.internal.address**
```
balhar-jakub: This is actually configs_server_internal_address

So most probably the consumption will be:

`components.gateway.server.internal.address`

components:
    gateway:
        server:
            internal:
                address

Or for single-service deployment ZWE_configs_internal_discovery_address

so - `components.apiml.internal.discovery.address`

docs/user-guide/address-network-requirements.md
```diff
@@ -11,9 +11,15 @@ For more information about variable names in the following table, see the [Zowe
...
 
 Most Components of Zowe are HTTPS servers. The ports of each and their default jobnames are listed below.
-The ports can be customized for each component by editing the value of `components.<component-name>.port` within the Zowe YAML file.
+The ports can be customized for each component by editing the value of `components.component-name.port` within the Zowe YAML file.
 Each Jobname has a default prefix of ZWE1, but that can be customized via the `zowe.job.prefix` value in the Zowe YAML file. The ports used are different between single-service API ML deployment and multi-service API ML deployment.
 
+:::note
+**Internal and External Port Isolation**
+
+Zowe v3 allows you to assign distinct ports and IP addresses to internal and external listeners. This separation enables you to implement firewall rules that prevent external users from accessing sensitive internal management endpoints, even if the services are hosted on the same LPAR.
```
balhar-jakub: If you mention the ports here, do we document somewhere how to do that?

docs/user-guide/address-network-requirements.md
```diff
@@ -57,11 +63,60 @@ The Caching Service will use these additional ports if enabled (`components.cach
...
 
-If this default is not desired, it is recommended to use [TCPIP port assignment statements](https://www.ibm.com/docs/en/zos/2.5.0?topic=assignments-profiletcpip-port) to restrict the IP & ports of each server by their jobnames. The jobnames of each Zowe component is derived from the property `zowe.job.prefix` and `<component-suffix>` as shown in the table prior.
+:::note 
+**Using the Wildcard Address**
 
-When `zowe.job.prefix` is `ZWE1`, an example of port reservations with a fixed IP of `10.11.12.13` could be as follows with single-service API ML deployment:
+The address `0.0.0.0` acts as a wildcard, instructing Zowe to bind to all available network interfaces on the host. This is the default behavior and is recommended for environments where strict interface isolation is not required.
+:::
+
+### Binding to Specific or Multiple Addresses
```
balhar-jakub: There are two different things in play
 - One is about differentaiation between internal and external 

The other one is about binding to multiple different network interfaces. 

I am not sure if the heading is well selected, but I do understand that we somewhat mixed these mechanisms together.

docs/user-guide/address-network-requirements.md
janan07: I refactored this. Please review.

docs/user-guide/address-network-requirements.md
janan07: I did not find any documentation that shows how to do this, so I changed the language.

docs/user-guide/address-network-requirements.md
janan07: Ok. I tried to separate these two areas in my latest commit.

docs/user-guide/address-network-requirements.md
```diff
@@ -55,13 +61,69 @@ The Caching Service will use these additional ports if enabled (`components.cach
...
+
+The address `0.0.0.0` acts as a wildcard, instructing Zowe to bind to all available network interfaces on the host. This is the default behavior and is recommended for environments where strict interface isolation is not required.
+:::
+
+If you need to restrict Zowe to specific TCP/IP stacks or separate internal service traffic from external user traffic, use the following properties in zowe.yaml:
+
+* **zowe.network.server.listenAddresses**  
+Defines the specific IP address or a comma-separated list of addresses for the server to bind to.
+
+* **zowe.network.external.address**
```
balhar-jakub: It seems that for the actual external binding there are three things that work together:
- zowe.network.server.listenAddresses - This deinfes list of IP addresses e.g. Network interfaces that Zowe / API ML listens on
- zowe.externalPort - The main external port, this is just one number
- zowe.externalDomains - List of domains that are valid to use with the Zowe instance

docs/user-guide/address-network-requirements.md
```diff
@@ -55,13 +61,69 @@ The Caching Service will use these additional ports if enabled (`components.cach
...
+:::note Internal and External Listener Isolation
 
-If this default is not desired, it is recommended to use [TCPIP port assignment statements](https://www.ibm.com/docs/en/zos/2.5.0?topic=assignments-profiletcpip-port) to restrict the IP & ports of each server by their jobnames. The jobnames of each Zowe component is derived from the property `zowe.job.prefix` and `<component-suffix>` as shown in the table prior.
+Zowe v3 provides isolation between internal and external listeners, enabling you to implement firewall rules that prevent external users from accessing sensitive management endpoints. 
+:::  
 
-When `zowe.job.prefix` is `ZWE1`, an example of port reservations with a fixed IP of `10.11.12.13` could be as follows with single-service API ML deployment:
+**Example:**
+```yaml
+zowe:
```
balhar-jakub, line 96: I menitoned above, zowe.network.external.address is something I haven't found anywhere in the code, where does it come from?

docs/user-guide/address-network-requirements.md
janan07: Do I understand correctly then that these two parms:
zowe.externalPort - The main external port, this is just one number
zowe.externalDomains - List of domains that are valid to use with the Zowe instance

replace these:

* **zowe.network.external.address**  
Defines the address used for external listeners (end-user traffic).

* **components.apiml.internal.discovery.address** (used in single-service deployment)   
Defines the address used for internal listeners for service-to-service communication. In a multi-service deployment, use `components.gateway.server.internal.address`.

docs/user-guide/address-network-requirements.md
janan07: AI hallucination

janan07: I'll use this yaml reference to validate these parms going forward.
https://docs.zowe.org/stable/appendix/zowe-yaml-configuration/

docs/user-guide/address-network-requirements.md
```diff
@@ -71,8 +128,11 @@ When `zowe.job.prefix` is `ZWE1`, an example of port reservations with a fixed I
    7557 TCP ZWE1SZ BIND 10.11.12.13 ; Zowe ZSS
 ```
 
+**Multi-Service Deployment**
+
+In a multi-service deployment, each component runs in its own address space with a unique jobname suffix.
 
-With multi-service deployment, it could be as follows:
+**Example:**
```
janan07, line 146: Are all of these binding addresses accurate for both internal and external communication?

docs/user-guide/address-network-requirements.md
balhar-jakub: That helps, also this file https://github.com/zowe/zowe-install-packaging/blob/v3.x/staging/example-zowe.yaml should contain the most

docs/user-guide/address-network-requirements.md
```diff
@@ -55,13 +61,64 @@ The Caching Service will use these additional ports if enabled (`components.cach
...
+
+If you need to restrict Zowe to specific TCP/IP stacks or separate internal service traffic from external user traffic, use the following properties in zowe.yaml:
 
-Zowe's servers by default use the TCP IP address `0.0.0.0` which assigns the servers to be available on all network interfaces available to the jobs.
+* **zowe.network.server.listenAddresses**  
+Defines the specific IP address or a comma-separated list of addresses for the server to bind to.
 
-If this default is not desired, it is recommended to use [TCPIP port assignment statements](https://www.ibm.com/docs/en/zos/2.5.0?topic=assignments-profiletcpip-port) to restrict the IP & ports of each server by their jobnames. The jobnames of each Zowe component is derived from the property `zowe.job.prefix` and `<component-suffix>` as shown in the table prior.
+* **zowe.externalPort**  
+Specifies the primary port number used for external user access to Zowe services. This is a single integer that defines the entry point for end-user traffic, typically routed through the API Gateway.
```
balhar-jakub: I would not add the routed through the API Gateway

balhar-jakub: Also I would add the fact that this is the port that needs to be exposed to the external traffic and as such requires opening of the firewall.

docs/user-guide/address-network-requirements.md
```diff
@@ -55,13 +61,64 @@ The Caching Service will use these additional ports if enabled (`components.cach
...
+
+* **zowe.externalDomains** (used in single-service deployment)   
+Specifies authorized domain names or hostnames that are valid for use with the Zowe instance. This property helps define the network boundaries and is used for security validations, such as ensuring that the Zowe instance only accepts requests directed to these specific domains.
```
balhar-jakub: This statement is not terribly imprecise, but not exactly precise either. 

This property specifies domains through which Zowe expects to be contacted. It is not per se part of any security infrastructure. 

The actual routing happens in different layers of the protocol, but for certain areas such as API Catalog transformations we need to have an information about what the external clients expect. We also use it to validate some things around certificates. 

I probably would not mention it here, but I was sharing it with you, so you have a better idea on how the process actually work as technically speaking from routing infrastructure point of view, this property is not really used.

janan07: Updated the language.

docs/user-guide/address-network-requirements.md
janan07: Removed Gateway mention.

docs/user-guide/address-network-requirements.md
janan07: The current draft has these three parms:

* **zowe.network.server.listenAddresses**  
Defines the specific IP address or a comma-separated list of addresses for the server to bind to.

* **zowe.externalPort**  
Specifies the primary port number used for external user access to Zowe services. This value defines the entry point for end-user traffic.

* **zowe.externalDomains** (used in single-service deployment)   
Specifies authorized domain names or hostnames through which Zowe expects to be contacted.

docs/user-guide/address-network-requirements.md
```diff
@@ -55,13 +61,63 @@ The Caching Service will use these additional ports if enabled (`components.cach
...
+* **zowe.network.server.listenAddresses**  
+Defines the specific IP address or a comma-separated list of addresses for the server to bind to.
+
+* **zowe.externalPort**  
+Specifies the primary port number used for external user access to Zowe services. This value defines the entry point for end-user traffic.
+
+* **zowe.externalDomains** (used in single-service deployment)   
+Specifies authorized domain names or hostnames through which Zowe expects to be contacted. 
+
+:::note Internal and External Listener Isolation
```
balhar-jakub: The removal of the information about the internal port setup is intentional?

janan07: The content is still included, just not as a note.

docs/user-guide/address-network-requirements.md
janan07: Would it be accurate to change these internal addresses to `127.0.0.1`

docs/user-guide/address-network-requirements.md
```diff
@@ -11,9 +11,15 @@ For more information about variable names in the following table, see the [Zowe
...
 
 Most Components of Zowe are HTTPS servers. The ports of each and their default jobnames are listed below.
-The ports can be customized for each component by editing the value of `components.<component-name>.port` within the Zowe YAML file.
+The ports can be customized for each component by editing the value of `components.component-name.port` within the Zowe YAML file.
 Each Jobname has a default prefix of ZWE1, but that can be customized via the `zowe.job.prefix` value in the Zowe YAML file. The ports used are different between single-service API ML deployment and multi-service API ML deployment.
 
+:::note
+**Internal and External Port Isolation**
+
+Zowe v3 provides isolation between internal and external listeners, enabling you to implement firewall rules that prevent external users from accessing sensitive management endpoints. For details, see [Binding to Specific or Multiple Addresses](#binding-to-specific-or-multiple-addresses) later in this article.
```
balhar-jakub: Didn't we already discuss this one? 

This statement is not correct - enabling you to implement firewall rules that prevent external users from accessing sensitive management endpoints

Honestly this whole part is a bit of mess - Zowe v3 provides isolation between internal and external listeners, enabling you to implement firewall rules that prevent external users from accessing sensitive management endpoints.

Zowe exposes internal and external port for the traffic. The internal port is used for the services to register themselves to the Zowe API Mediation Layer, the external port is used to access Zowe and all the onboarded services. The external port is also the correct port to use for opening firewalls and setting up firewall rules.

docs/user-guide/address-network-requirements.md
```diff
@@ -55,13 +61,73 @@ The Caching Service will use these additional ports if enabled (`components.cach
 | 7601        | zowe.components.caching-service.storage.infinispan.jgroups.keyExchange.port | The port at which the key server in Infinispan is listening. If the port is not available, the next port is probed, up to port+5. Used by the key server (server) to create an SSLServerSocket and by clients to connect to the key server. |
 | 7600        | zowe.components.caching-service.storage.infinispan.jgroups.port | Bind port for the socket that is used to form an Infinispan cluster. |
 
-## IP Addresses
+## Network Configuration
+
+Configure how Zowe identifies and binds to network interfaces on your system to ensure accessibility for users while maintaining secure internal communication between services.
```
balhar-jakub: The network interfaces are not primarily for securing internal communication between services, if at all.

The primary use case here is that for some of the systems there are multiple Network Interface Configuration available and the Zowe is intended to be used on just one of them. 

So something that is more correct would be:
Configure how Zowe identifies and binds to network interfaces on your system. The default behavior binds to all interfaces, but it is possible to configure binding to just specific subset of available Network Interfaces (NIC)

docs/user-guide/address-network-requirements.md
```diff
@@ -55,13 +61,73 @@ The Caching Service will use these additional ports if enabled (`components.cach
...
+  # List of valid domains for the Zowe instance
+  externalDomains: 
+    - "zowe.example.com"
+    - "api.internal.org"
 
-If this default is not desired, it is recommended to use [TCPIP port assignment statements](https://www.ibm.com/docs/en/zos/2.5.0?topic=assignments-profiletcpip-port) to restrict the IP & ports of each server by their jobnames. The jobnames of each Zowe component is derived from the property `zowe.job.prefix` and `<component-suffix>` as shown in the table prior.
+components:
+  apiml:
+    internal:
+      # Defines the address for service-to-service communication
```
balhar-jakub: # Define the address for the communication from z/OS services to the endpoint managing the registration of the services.

docs/user-guide/address-network-requirements.md
```diff
@@ -55,13 +61,73 @@ The Caching Service will use these additional ports if enabled (`components.cach
...
+
+Zowe v3 provides isolation between internal and external listeners, enabling the network administrator or system programmer to implement firewall rules that prevent external users from accessing sensitive management endpoints. 
+:::  
+
+**Example:**
+```yaml
+zowe:
+  network:
+    server:
+      # Binds to both local and external interfaces
```
balhar-jakub: This, while correct, may be misleading. 

The concept of local interfaces is more off z/OS thing. I would rather use two IP addresses e.g. 10.11.12.13 and 11.12.13.14

and as the explanation - Binds to two of the Network interfaces available on the system.

docs/user-guide/address-network-requirements.md
```diff
@@ -55,13 +61,73 @@ The Caching Service will use these additional ports if enabled (`components.cach
...
+      listenAddresses: "127.0.0.1, 10.11.12.13"
+  
+  # The main external port for end-user traffic
```
balhar-jakub: Maybe better to use incoming or client traffic? The end-user may be another service, end-user points more into direction of person being involved.

docs/user-guide/address-network-requirements.md
balhar-jakub: Network Interfaces are not about internal vs. external communication, at least not per se. 

With this stated please don't use the 127.0.0.1 to showcase this as it is about multiple Network Interfaces available on specific z/OS system (you may also hear about the term multiple TCP/IP Stacks)

So the 10.11.12.13 is totally fine

janan07: Ok. Will address this after standup.

docs/user-guide/address-network-requirements.md
janan07: I've refactored this section. Please review.

balhar-jakub: I see the same text within this paragraph, is it lost somewhere?

balhar-jakub: This comment was specifically for this paragraph it is in. I see that the section below was changed.

docs/user-guide/address-network-requirements.md
```diff
@@ -55,13 +61,74 @@ The Caching Service will use these additional ports if enabled (`components.cach
...
-## IP Addresses
+## Network Configuration
+
+Configure how Zowe identifies and binds to network interfaces on your system. The default behavior binds to all interfaces, but it is possible to configure binding to just a specific subset of available Network Interfaces (NIC).
+
+### Interface Binding and Traffic Isolation
+
+By default, Zowe binds to the wildcard address `0.0.0.0`, making Zowe services available on all network interfaces. This configuration is recommended for environments where strict interface isolation is not required.
+
+Zowe V3 utilizes a dual-listener architecture to separate internal traffic from user traffic. If you need to restrict Zowe to specific TCP/IP stacks or separate internal service traffic, use the following properties in zowe.yaml:
```
balhar-jakub: the wording - utilizes a dual-listener architecture ... is really strange

balhar-jakub: What about Zowe uses two listeners to separate internal traffic from external.

docs/user-guide/address-network-requirements.md
```diff
@@ -55,13 +61,74 @@ The Caching Service will use these additional ports if enabled (`components.cach
...
+
+* **Internal Port**  
+Used exclusively for Zowe services to register themselves with the API Mediation Layer and for service-to-service discovery. This traffic should be restricted to the local or internal network.
```
balhar-jakub: The correct statement would be:
Used for services to register themselves with the API Mediation Layer. This traffic is typically restricted to the internal network.

Note: There are extenders that register off-platform services, in that case it needs to be available to these services as well.

docs/user-guide/address-network-requirements.md
```diff
@@ -55,13 +61,74 @@ The Caching Service will use these additional ports if enabled (`components.cach
...
+* **External Port**  
+The primary gateway for all external user access (Web UI, CLI, VS Code) and onboarded services. This is the port used for opening firewalls and defining external access rules.
```
balhar-jakub: Why is the gateway here? Why not just use port or point of acces. 
Gateway have also other meanings from the architecture point of view. Including the fact that one of the services is named this way.

docs/user-guide/address-network-requirements.md
janan07: Language refactored

docs/user-guide/address-network-requirements.md
janan07: updated.

docs/user-guide/address-network-requirements.md
janan07: I've updated the note as the following:

"Zowe v3 supports the use of separate internal and external listeners to isolate service-to-service communication from external traffic. For information on how to configure these listeners and their associated ports, see [Interface Binding and Traffic Isolation](#interface-binding-and-traffic-isolation) later in this article."

docs/user-guide/address-network-requirements.md
janan07: Language has been refactored to refer to the primary access point for all external user access.

docs/user-guide/address-network-requirements.md
janan07: Updated the Internal Ports description

docs/user-guide/address-network-requirements.md
janan07: Thanks for the clarification. Using the hypothetical addresses: 10.11.12.13 and 11.12.13.14

docs/user-guide/address-network-requirements.md
janan07: Definition updated

docs/user-guide/address-network-requirements.md
janan07: updated

docs/user-guide/address-network-requirements.md
balhar-jakub: I have the issue with service-to-service term, this would imply far wider intentional usage and that is not correct. 

This internal port is intended only to be used for registration of the services to the API ML

janan07: Gotcha. I saw this in one of your previous comments. Noted.

docs/user-guide/address-network-requirements.md
```diff
@@ -11,9 +11,15 @@ For more information about variable names in the following table, see the [Zowe
...
 
 Most Components of Zowe are HTTPS servers. The ports of each and their default jobnames are listed below.
-The ports can be customized for each component by editing the value of `components.<component-name>.port` within the Zowe YAML file.
+The ports can be customized for each component by editing the value of `components.component-name.port` within the Zowe YAML file.
 Each Jobname has a default prefix of ZWE1, but that can be customized via the `zowe.job.prefix` value in the Zowe YAML file. The ports used are different between single-service API ML deployment and multi-service API ML deployment.
 
+:::note
+**Internal and External Port Isolation**
+
+Zowe v3 supports the use of separate internal and external listeners to isolate service-to-service communication from external traffic. For information on how to configure these listeners and their associated ports, see [Interface Binding and Traffic Isolation](#interface-binding-and-traffic-isolation) later in this article.
```
balhar-jakub: service-to-service communication should not be used here as it would imply technical usage that is not intended. 

Better explanation/wording of this paragraph would be something like:
Zowe supports external port for incoming traffic and it's routing and internal port for registration of the services primarily running on z/OS via internal channel. For information on how to configure these listeners and their associated ports, see [Interface Binding and Traffic Isolation](#interface-binding-and-traffic-isolation) later in this article.

balhar-jakub (approved):

PR #5043, opened 2026-05-27
Add details on when password setup for user is required

balhar-jakub:
Co-Authored-By: Hermes (Various Models)

Depending on details such as authentication providers, it is possible that even when password is not directly user, the user needs to have a password.
This PR outlines when exactly is this happening

docs/user-guide/address-authentication-requirements.md
docs/user-guide/authenticating-with-client-certificates.md
docs/user-guide/configure-uss.md

### DIFF
--- a/docs/user-guide/configure-uss.md
+++ b/docs/user-guide/configure-uss.md
@@ -38,9 +38,10 @@ To ensure successful authentication and operation through Zowe, verify that user
 
 * **Password Status**  
 Password requirements depend on your authentication provider:
+
   * **z/OSMF provider**: The user ID must have a password or passphrase enabled (not NOPASSWORD), even if using certificate-based authentication.
-  * **SAF provider (e.g., RACF, ACF2, Top Secret)**: Users authenticating with client certificates are not required to have a password for login. However, a password is required if the user needs PassTicket-based access to z/OSMF or downstream services. Suspended users cannot authenticate with client certificates in either case.
-  * For Zowe service IDs (e.g., `ZWESVUSR`, `ZWESIUSR`), ensure the password is set to non-expire to avoid service interruptions, regardless of provider.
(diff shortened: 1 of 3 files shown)


docs/user-guide/configure-uss.md
```diff
@@ -37,7 +37,10 @@ For information about OMVS segments, see the article _The OMVS segment in user p
 To ensure successful authentication and operation through Zowe, verify that user profiles meet the following criteria:
 
 * **Password Status**  
-The user ID must have a password or passphrase enabled (not NOPASSWORD), even if using certificate-based authentication. For Zowe service IDs, ensure the password is set to non-expire to avoid service interruptions.
+Password requirements depend on your authentication provider:
+  * **z/OSMF provider**: The user ID must have a password or passphrase enabled (not NOPASSWORD), even if using certificate-based authentication.
+  * **SAF provider (e.g., RACF, ACF2, Top Secret)**: Users authenticating with client certificates are not required to have a password for login. However, a password is required if the user needs PassTicket-based access to z/OSMF or downstream services. Suspended users cannot authenticate with client certificates in either case.
+  * For Zowe service IDs (e.g., `ZWESVUSR`, `ZWESIUSR`), ensure the password is set to non-expire to avoid service interruptions, regardless of provider.
```
janan07: please verify that `non-expire` is an actual value.

balhar-jakub: It is not an actual value as this is not example, but it is a statement that is also used in the z/OS documentation by IBM

"rather than using the NOEXPIRED operand to assign non-expired passwords"

But if we want to go into more details and the correct parameters, then the correct example is NOINTERVAL which effectively sets that the password does not expire. 

"User IDs that have NOINTERVAL specified as their password interval. Value: Identifies users who are not required to change their passwords."

janan07 (approved):

PR #5094, opened 2026-07-17
Document cors defaults parms / security overhaul

janan07:
Describe your pull request here: Document cors defaults parms / security overhaul


docs/user-guide/api-mediation/configuration-cors.md
```diff
@@ -1,32 +1,79 @@
...
 
-:::info Role: system programmer
+:::info Required Role: system programmer
 :::
 
-As a system programmer, you can enable the Gateway to terminate CORS requests for itself and also for routed services. By default, Cross-Origin Resource Sharing (CORS) handling is disabled for Gateway routes `gateway/api/v1/**` and for individual services. After enabling the feature as stated in the following procedure, API Gateway endpoints start handling CORS requests. Individual services can control whether they want the Gateway to handle CORS for them through the [Custom Metadata](../../extend/extend-apiml/onboard-spring-boot-enabler.md#custom-metadata) parameters.
+You can enable the Gateway to terminate CORS requests for itself and also for routed services. By default, Cross-Origin Resource Sharing (CORS) handling is disabled for Gateway routes `gateway/api/v1/**` and for individual services. Once enabled, API Gateway endpoints handle CORS requests according to your global configuration, and individual services can delegate CORS handling to the Gateway using per-service [Custom Metadata](../../extend/extend-apiml/onboard-spring-boot-enabler.md#custom-metadata) CORS parameters.
```
pablocarle, line 6: maybe it should specify what it means to have CORS disabled. In CORS disabled mode it is at its most strict mode in normal web applications.
In the case of API ML Gateway, disabling CORS means respecting Gateway's defaults which are already covered in the docs and can be customized via zowe.yaml settings.

pablocarle: The difference of course is that in the disabled mode, Gateway will not follow any CORS directive coming from the service.

docs/user-guide/api-mediation/configuration-cors.md
```diff
@@ -1,32 +1,79 @@
...
+## Preflight Request Management
+
+When CORS is enabled, the Gateway automatically answers preflight (OPTIONS) requests on behalf of the service, replying with the following default CORS headers
 
-Additionally, the Gateway handles the preflight requests on behalf of the service when CORS is enabled in Custom Metadata, replying with CORS headers:
 - `Access-Control-Allow-Methods: GET,HEAD,POST,PATCH,DELETE,PUT,OPTIONS`
+   This setting can be overridden globally using `components.gateway.apiml.service.corsAllowedMethods` in zowe.yaml. 
 - `Access-Control-Allow-Headers: origin, x-requested-with`
 - `Access-Control-Allow-Credentials: true`
-- `Access-Control-Allow-Origin: *`
```
balhar-jakub, line 19: I believe that the origin remains, only the default changed. So we probably should not remove it.

docs/extend/extend-apiml/onboard-direct-eureka-call.md
```diff
@@ -207,7 +207,19 @@ HttpClient employs a special entity to manage access to HTTP connections called
...
 
 * **apiml.service.corsEnabled**  
-When this parameter is set to `true`, CORS is enabled on the service level for all service routes. The same parameter can also be set on the service level, by providing the parameter as `customMetadata` as shown in the [Custom Metadata](./custom-metadata.md).
+  When this parameter is set to `true`, CORS is enabled on the service level for all service routes. The same parameter can also be set on the service level by providing the parameter as `customMetadata` as shown in [Custom Metadata](./custom-metadata.md).
+
+* **customMetadata.apiml.corsAllowedOrigins**  
+  (Optional) Specify which origins are accepted by the Gateway during CORS handling for this service. If not set, this value falls back to the Gateway's global `corsDefaultAllowedOrigins` default (`https://${hostname}:${port}`). Inclusion of other allowed orgins is additive to the Gateway defaults rather than replacing these defaults.
```
balhar-jakub, line 213: Typo - orgins instead of origins.

docs/user-guide/api-mediation/configuration-cors.md
```diff
@@ -1,32 +1,79 @@
...
+* **corsDefaultAllowedOrigins**  
+Specifies the allowed origins for CORS requests across the Gateway. This parameter does not default to a wildcard (`*`). Instead, to ensure full security, this parameter defaults strictly to the Gateway's own host and port. This parameter can be customized by adding alternative origins or explicitly setting it to `*` if open access is required. Individual services can also override this value using the `customMetadata.apiml.corsAllowedOrigins` parameter.  
+**Default:** `https://${hostname}:${port}`
+
+* **corsDefaultAllowedHeaders**  
+Defines which HTTP headers are accepted during a CORS request. By default, all headers are permitted, but this can be restricted to a specific comma-separated list of headers to match your enterprise security requirements.  
+**Default:** `*`
+
+## Per-Service Custom Metadata CORS Parameters
+Service administrators can fine-tune CORS behavior on a per-service basis. During service registration, you can pass specific CORS configuration parameters via Custom Metadata. For more information, see [Customizing Matadata (optional)](../../extend/extend-apiml/custom-metadata.md).
```
balhar-jakub: Typo - Customizing Matadata instead of Customizing Metadata

docs/user-guide/api-mediation/configuration-cors.md
```diff
@@ -1,32 +1,79 @@
...
-When the Gateway handles CORS on behalf of the service, the Gateway sanitizes the following defined headers from the communication (upstream and downstream) in the following comma -separated list:
-```
-Access-Control-Request-Method,Access-Control-Request-Headers,Access-Control-Allow-Origin,Access-Control-Allow-Methods,Access-Control-Allow-Headers,Access-Control-Allow-Credentials,Origin
-```
 
-The resulting request to the service is not a CORS request. No additional specification of the service is required. The list can be overridden by specifying a different comma-separated list in the property `components.gateway.apiml.service.ignoredHeadersWhenCorsEnabled` in `zowe.yaml`.
+## Gateway-Managed CORS Handling
+When the Gateway handles CORS on behalf of a service, the Gateway strips specific CORS handshake headers from the upstream and downstream communication loop to prevent them from interfering with upstream or downstream components.
+
+## Header Removal
```
balhar-jakub: Shouldn't this header be a deeper embedding e.g. ### instead of ##

docs/extend/extend-apiml/custom-metadata.md
```diff
@@ -49,40 +46,58 @@ Specifies whether all operations can be retried for this service. The default va
...
+  :::
+
+* **customMetadata.apiml.corsAllowCredentials**  
+(Optional) Configure whether the browser should include credentials (such as cookies, authorization headers, or TLS client certificates) in CORS requests. Note that while the Gateway-level default is hardcoded to `true`, the per-service default for this parameter is `false`.  
+  :::note
+  If you use the Spring enabler, use the following parameter name:
+  `apiml.service.customMetadata.apiml.corsAllowCredentials`
+  :::
+
+* **customMetadata.apiml.corsAllowedMethods**
```
balhar-jakub: We probably missed the property customMetadata.apiml.corsAllowedMethods here.

docs/user-guide/api-mediation/configuration-cors.md
```diff
@@ -1,32 +1,83 @@
...
+You can enable the Gateway to terminate CORS requests for itself and also for routed services. By default, Cross-Origin Resource Sharing (CORS) handling is disabled for Gateway routes `gateway/api/v1/**` and for individual services. Once enabled, API Gateway endpoints handle CORS requests according to your global configuration, and individual services can delegate CORS handling to the Gateway using per-service [Custom Metadata](../../extend/extend-apiml/onboard-spring-boot-enabler.md#custom-metadata) CORS parameters. 
 
-When the Gateway handles CORS on behalf of the service, the Gateway sanitizes the following defined headers from the communication (upstream and downstream) in the following comma -separated list:
-```
-Access-Control-Request-Method,Access-Control-Request-Headers,Access-Control-Allow-Origin,Access-Control-Allow-Methods,Access-Control-Allow-Headers,Access-Control-Allow-Credentials,Origin
-```
+When CORS handling is disabled for a service, the Gateway does not follow or apply any CORS directives defined by that service. Instead, request handling relies entirely on the Gateway's global defaults, which are configured via settings in zowe.yaml.
```
pablocarle: when CORS handling is disabled for API ML

docs/user-guide/api-mediation/configuration-cors.md
pablocarle: in both cases the request headers do not reach the service, these are response headers though.

if Gateway handles CORS on behalf of the service: if the service provides settings, use them, otherwise global defaults.
if Gateway has CORS handling disabled: global gateway defaults, ignore service requirements.

Allow-Origin: * is not the default nor allowed.

docs/user-guide/api-mediation/configuration-cors.md
janan07: revised

docs/user-guide/api-mediation/configuration-cors.md
balhar-jakub: Just a question, do I read correctly that in case gatewayEnabled is true, the CORS related headers are always stripped

In case it is false, they remain?

pablocarle: CORS request headers have to be stripped in every case, in particular Origin header, otherwise the southbound service may not match to gateway's host.
The difference relies on whether gateway will take service's overrides or not

docs/user-guide/api-mediation/configuration-cors.md
```diff
@@ -1,32 +1,83 @@
...
+
+:::tip
+When CORS handling is enabled, you can override this default list of stripped headers by specifying a custom comma-separated list via the `components.gateway.apiml.service.ignoredHeadersWhenCorsEnabled` property in `zowe.yaml`.
+:::
+
+## Preflight Request Management
+
+When CORS handling is enabled, the Gateway automatically answers preflight (`OPTIONS`) requests on behalf of the service, replying with the following default CORS headers:
 
-Additionally, the Gateway handles the preflight requests on behalf of the service when CORS is enabled in Custom Metadata, replying with CORS headers:
```
pablocarle, line 15: I think this needs to be updated, the preflight is always handled by gateway, the only difference is whether Gateway will take service-provided overrides (with CORS enabled on Gateway) or use the global defaults (with disabled)
In the first scenario, if the service does not provide overrides, the global defaults are used.
The global defaults can be changed for the Gateway in zowe.yaml.

docs/user-guide/api-mediation/configuration-cors.md
janan07: revised

docs/user-guide/api-mediation/configuration-cors.md
balhar-jakub: So to validate I understand the whole piece:

Regardless of whether the corsEnabled is true or false. The API ML strips Access-Control-Allow-* Headers and then the discussion is what happens with respect to addition of CORS headers for the downstream service. 

If corsEnabled is false, there are no CORS headers added. 

If corsEnabled is true, there are default CORS headers added and that is it unless the service provides the additional headers information

pablocarle: Only request headers are stripped, so `Origin` and `Access-Control-Request-*` headers (for preflight OPTIONS requests)
Response headers are untouched (`Access-Control-Allow-*`)
Since request headers are removed, the service always receives a non-CORS request.

The only difference is that with corsEnabled `true`, Gateway may respond with service-specific response headers instead of the defaults. Enphasis on the "may", since it requires that the service provides the correct metadata values.

balhar-jakub: Let me then reiterate my understanding. I will try to show the three options - 
1) corsEnabled: false
2) corsEnabled: true, customMetadata.apiml.corsEnabled: true
3) corsEnabled: true, customMetadata.apiml.corsEnabled: false

The flows

1) The Request comes with the headers Origin, Access-Control-Request-Method, Access-Control-Request-Headers 
    The API ML strips all these headers from this request
    The API ML calls the downstream service without Access-Control-Request-Method, Access-Control-Request-Headers
    The service may set up Response headers e.g. Access-Control-Allow-* 
    If these headers are set up, this is what the API ML returns and it does not touch them. 

2) The Request comes with the headers Origin, Access-Control-Request-Method, Access-Control-Request-Headers 
    The API ML strips all these headers from this request
    The API ML calls the downstream service with Access-Control-Request-Method, Access-Control-Request-Headers set based on the defaults plus the specific informations provided in customMetadata by service
    The service may set up Response headers e.g. Access-Control-Allow-* 
    If these headers are set up, this is what the API ML returns and it does not touch them. 

3) The Request comes with the headers Origin, Access-Control-Request-Method, Access-Control-Request-Headers 
    The API ML strips all these headers from this request
    The API ML calls the downstream service with Access-Control-Request-Method, Access-Control-Request-Headers set based on the defaults only
    The service may set up Response headers e.g. Access-Control-Allow-* 
    If these headers are set up, this is what the API ML returns and it does not touch them.

pablocarle: Not exactly.
In all cases CORS is resolved by Gateway, the service does not see a CORS request at all (since the headers were removed)

The actual requests get to the service (with CORS request headers removed). But, before the actual request there is a preflight request that browsers typically do, these are HTTP OPTIONS requests.
The preflight request is resolved by Gateway independently of whether the CORS handling is enabled or not (these requests are never sent to the service) based on the CORS settings

So in the 3 scenarios above I'd update that the `Access-Control-Allow-*` headers are set by Gateway, not by the service.

pablocarle: Another clarifycation, in the scenarios described above, if the request gets to the service, since the CORS headers were removed, it's not expected that the service will set any CORS response headers

docs/user-guide/api-mediation/configuration-cors.md
```diff
@@ -1,32 +1,99 @@
...
 
-:::info Role: system programmer
+:::info Required Role: system programmer
 :::
 
-As a system programmer, you can enable the Gateway to terminate CORS requests for itself and also for routed services. By default, Cross-Origin Resource Sharing (CORS) handling is disabled for Gateway routes `gateway/api/v1/**` and for individual services. After enabling the feature as stated in the following procedure, API Gateway endpoints start handling CORS requests. Individual services can control whether they want the Gateway to handle CORS for them through the [Custom Metadata](../../extend/extend-apiml/onboard-spring-boot-enabler.md#custom-metadata) parameters.
+You can enable the Gateway to terminate CORS requests for itself and also for routed services. By default, Cross-Origin Resource Sharing (CORS) handling is disabled for Gateway routes `gateway/api/v1/**` and for individual services. Once enabled, API Gateway endpoints handle CORS requests according to your global configuration, and individual services can delegate CORS handling to the Gateway using per-service [Custom Metadata](../../extend/extend-apiml/onboard-spring-boot-enabler.md#custom-metadata) CORS parameters.
```
balhar-jakub, line 6: **Factual error: CORS default state.** The statement "CORS handling is **disabled** for Gateway routes `gateway/api/v1/**` and for individual services" is incorrect. The current default in both `gateway-service/src/main/resources/application.yml` and `apiml/src/main/resources/application.yml` is `corsEnabled: true`. CORS is enabled by default; only the origin list is restricted to the gateway's own hostname. Suggested correction:

```
By default, Cross-Origin Resource Sharing (CORS) handling is **enabled** with restrictive defaults: only the Gateway's own URL is permitted as a request origin. To expand the list of allowed origins, use the configuration properties described below.
```

docs/user-guide/api-mediation/configuration-cors.md
```diff
@@ -1,32 +1,99 @@
...
+
+* **CORS Disabled**  
+When CORS handling is disabled, the Gateway strictly uses the global defaults and ignores any service-provided overrides.
+
+### Default CORS Response Headers
+
+Unless overridden globally in `zowe.yaml` or by service-specific metadata (when enabled), the Gateway replies with the following default CORS response headers:  
+
+  * **`Access-Control-Allow-Methods: GET,HEAD,POST,PATCH,DELETE,PUT,OPTIONS`**  
+    Can be overridden globally using `components.gateway.apiml.service.corsDefaultAllowedMethods` in `zowe.yaml`.
```
balhar-jakub, line 45: **Property name typo.** The doc references `components.gateway.apiml.service.corsDefaultAllowedMethods`, but the actual property name in code (`ConnectionsConfig.java` line 100) is `apiml.service.corsAllowedMethods`. There is no `corsDefaultAllowedMethods` property. Should be `corsAllowedMethods`.

docs/user-guide/api-mediation/configuration-cors.md
```diff
@@ -1,32 +1,99 @@
...
+  * **`Access-Control-Allow-Headers: origin, x-requested-with`**  
+    Can be overridden globally using `components.gateway.apiml.service.corsDefaultAllowedHeaders` in `zowe.yaml`.
```
balhar-jakub, line 47: **Default value mismatch.** This lists `Access-Control-Allow-Headers: origin, x-requested-with` as the default, but the actual code default (`ConnectionsConfig.java` line 106) for `corsDefaultAllowedHeaders` is `*`. Either:
- Change the code default to `origin, x-requested-with` for tighter security (a security improvement), or
- Change the doc to show `*` as the actual default.

The same fix should also be made in `docs/appendix/zowe-yaml-configuration.md` line 597-599.

docs/user-guide/api-mediation/configuration-cors.md
```diff
@@ -1,32 +1,99 @@
...
+
+### Effect of the CORS Enabled/Disabled State
+
+Enabling or disabling CORS handling purely determines how the Gateway constructs the CORS response headers for the preflight request:
+
+* **CORS Enabled**  
+When CORS handling is enabled, the Gateway may respond with service-specific CORS response headers, provided that the service supplies the correct metadata values to override the defaults. If the service does not provide these overrides, the Gateway falls back to the global defaults.
+
+* **CORS Disabled**  
+When CORS handling is disabled, the Gateway strictly uses the global defaults and ignores any service-provided overrides.
```
balhar-jakub, line 38: **Missing context: this section needs a note about the relationship between gateway-level and per-service configuration.** When `corsEnabled=false`, the gateway applies defaults to `/` (all paths) — but per-service overrides via `apiml.corsAllowedOrigins` are still ignored. This is consistent with the `CorsUtils.registerDefaultCorsConfiguration` logic (lines 119-135 of `common-service-core/.../CorsUtils.java`).

Suggest adding: "When CORS is disabled, all endpoints still receive the Gateway's default CORS response headers — the difference is that no `apiml.cors*` metadata from registered services is consulted."

docs/user-guide/api-mediation/configuration-cors.md
```diff
@@ -1,32 +1,99 @@
...
+
+Unless overridden globally in `zowe.yaml` or by service-specific metadata (when enabled), the Gateway replies with the following default CORS response headers:  
+
+  * **`Access-Control-Allow-Methods: GET,HEAD,POST,PATCH,DELETE,PUT,OPTIONS`**  
+    Can be overridden globally using `components.gateway.apiml.service.corsDefaultAllowedMethods` in `zowe.yaml`.
+  * **`Access-Control-Allow-Headers: origin, x-requested-with`**  
+    Can be overridden globally using `components.gateway.apiml.service.corsDefaultAllowedHeaders` in `zowe.yaml`.
+  * **`Access-Control-Allow-Credentials: true`**
+  * **`Access-Control-Allow-Origin: https://${hostname}:${port}`**    
+
```
balhar-jakub, line 50: **Hardcoded value not configurable.** The doc says `Access-Control-Allow-Credentials: true` is a default, but in `CorsUtils.registerDefaultCorsConfiguration` (current code, post-PR #4848), `setAllowCredentials(true)` is hardcoded for gateway defaults — it cannot be overridden via configuration today. Recommend either documenting this as a constraint, or making it configurable.

docs/user-guide/api-mediation/configuration-cors.md
```diff
@@ -1,32 +1,99 @@
...
+* **`apiml.corsAllowedOrigins`**  
+Specifies the allowed origins for the service.
+* **`apiml.corsAllowedHeaders`**  
+Restricts accepted HTTP headers for the service.
+* **`apiml.corsAllowCredentials`**  
+Configures credential sharing capabilities.
+* **`apiml.corsAllowedMethods`**  
+Sets the allowed HTTP methods for the service.
+
+## Enabling CORS Handling
```
balhar-jakub, line 80: **Missing security additions from PR #4823 (July 2026).** Since this PR is titled "security overhaul," it should also document:

1. **`security.secFetch.enabled`** and **`security.secFetch.safeNavigationDestinations`** — opt-in token-free CSRF protection using the browser-supplied `Sec-Fetch-Site` header. Default is disabled. When enabled, blocks cross-site requests that don't pass CORS validation.

2. **Login endpoint content-type restriction** — the `/zaas/login` endpoint now requires `Content-Type: application/json` and rejects `application/x-www-form-urlencoded` (BREAKING CHANGE in PR #4823).

These are recent security hardenings that customers will need to know about. Without them, this overhaul is incomplete.

docs/user-guide/api-mediation/configuration-cors.md
```diff
@@ -1,32 +1,99 @@
...
+            corsEnabled: true
+            corsDefaultAllowedOrigins: "https://mytrusteddomain.com:7553"
+            corsDefaultAllowedHeaders: "Authorization,Content-Type"
+    ```
+
+4. Restart the Zowe instance to apply the new configuration.
   
-Requests through the Gateway now contain a CORS header. 
\ No newline at end of file
+Requests routed through the API Gateway will successfully process and append the appropriate CORS headers based on your global and per-service metadata definitions.
```
balhar-jakub, line 99: **Typo / duplicate period.** "based on your global and per-service metadata definitions. ." — double period before the closing line. Minor.

docs/appendix/zowe-yaml-configuration.md
```diff
@@ -592,6 +592,14 @@ User authorization is required to use the `IRR.RUSERMAP` resource within the `FA
  Specifies if the Gateway allows encoded characters to be part of URL requests redirected through the Gateway. Set to `true` to allow encoded characters to be part of URL requests. 
 - **apiml.service.corsEnabled**  
  Specifies if CORS are enabled in the API Gateway for Gateway routes `gateway/api/v1/**`. Set to `true` to enable CORS.
+- **apiml.service.corsDefaultAllowedOrigins**  
+  Specifies the default origins accepted by the Gateway during CORS handling when a per-service configuration is not specified  **Default:** `https://${hostname}:${port}`
+- **apiml.service.corsDefaultAllowedHeaders**
```
balhar-jakub, line 597: **Default value missing or incorrect.** This entry says the property specifies "the default HTTP headers allowed by the Gateway during a CORS request across all services when not explicitly overridden" but does not state the actual default. The actual code default (`ConnectionsConfig.java` line 106) is `*`. Should read something like:

**Default:** `*` (all headers)

If the intent is to tighten this default for security, that needs to be a code change in PR #4758 — not just a doc change.

docs/extend/extend-apiml/custom-metadata.md
```diff
@@ -49,40 +46,58 @@ Specifies whether all operations can be retried for this service. The default va
...
   For more information, refer to enabling CORS with Custom Metadata on the Gateway: [Customizing Cross-Origin Resource Sharing (CORS)](../../user-guide/api-mediation/configuration-cors.md).
+
+* **customMetadata.apiml.corsAllowedHeaders**  
+(Optional) Specify a comma-separated list of HTTP headers that are allowed during a CORS request to this service.    
+  :::note
+  If you use the Spring enabler, use the following parameter name:
+  `apiml.service.customMetadata.apiml.corsAllowedHeaders`
+  :::
+
+* **customMetadata.apiml.corsAllowCredentials**
```
balhar-jakub, line 86: **Default value clarification needed.** The note says "the per-service default is `false`, whereas the Gateway-level default is hardcoded to `true`" — this is correct, but the wording could mislead readers. The Gateway's hardcoded `true` applies to **gateway-default endpoints only** (when `corsEnabled=true` and the path matches `corsAllowedEndpoints`); it doesn't override per-service defaults. Suggest rewording: "If a service does not specify `apiml.corsAllowCredentials`, the per-service default is `false`. Note that the Gateway's default config (applied when CORS is enabled) hardcodes credentials to `true` for gateway routes; this setting controls behavior only for per-service CORS configurations."

docs/extend/extend-apiml/custom-metadata.md
```diff
@@ -49,40 +46,58 @@ Specifies whether all operations can be retried for this service. The default va
...
 * **customMetadata.apiml.gatewayPort**  
-Specifies the Gateway port used by the ZAAS Client configuration. The default value is `10010`. For more information about ZAAS Client, see [ZAAS Client](zaas-client.md).
-
+Specifies the Gateway port used by the ZAAS Client configuration. The default value is `10010`. For more information about ZAAS Client, see [ZAAS Client](zaas-client.md).  
   :::note
   If you use the Spring enabler, use the following parameter name:
   `apiml.service.customMetadata.apiml.gatewayPort`
   :::
 
 * **customMetadata.apiml.corsAllowedOrigins**
```
balhar-jakub, line 69: **Behavior of multiple origins should be clarified.** The doc lists `corsAllowedOrigins` but doesn't explain how multiple origins interact. In code (`CorsUtils.createCorsConfigurationForService`, lines 84-89): when a service sets `apiml.corsAllowedOrigins`, **those origins are additive** to the gateway's defaults — not a replacement. This means the effective allowed origins list = `corsDefaultAllowedOrigins` ∪ `corsAllowedOrigins`. Worth documenting explicitly so customers don't think they're overring the defaults.

docs/extend/extend-apiml/onboard-direct-eureka-call.md
```diff
@@ -207,7 +207,19 @@ HttpClient employs a special entity to manage access to HTTP connections called
 Specifies whether all operations can be retried for this service. The default value is `false`. The `false` value allows retries for only GET requests if a response code of `503` is returned. Setting this value to `true` enables retry requests for all methods, which return a `503` response code. Enabling retry can impact server resources resulting from buffering of the request body.
 
 * **apiml.service.corsEnabled**  
-When this parameter is set to `true`, CORS is enabled on the service level for all service routes. The same parameter can also be set on the service level, by providing the parameter as `customMetadata` as shown in the [Custom Metadata](./custom-metadata.md).
+  When this parameter is set to `true`, CORS is enabled on the service level for all service routes. The same parameter can also be set on the service level by providing the parameter as `customMetadata` as shown in [Custom Metadata](./custom-metadata.md).
+
+* **customMetadata.apiml.corsAllowedOrigins**
```
balhar-jakub, line 212: **Documentation duplication risk.** The same four parameters (`corsAllowedOrigins`, `corsAllowedHeaders`, `corsAllowCredentials`, `corsAllowedMethods`) are documented in two places: this file (direct Eureka call) and `custom-metadata.md` (Spring enabler). Recommend either:
- Linking from one file to the other and saying "see [Custom Metadata](custom-metadata.md) for descriptions", or
- Adding a single source-of-truth page that both files link to.

Otherwise, future changes will need to be made in two places and will inevitably drift.

docs/user-guide/api-mediation/configuration-cors.md
```diff
@@ -1,32 +1,99 @@
 # Customizing Cross-Origin Resource Sharing (CORS)
```
balhar-jakub, line 1: **Overall summary of review.**

This PR is a solid documentation overhaul that correctly documents the metadata keys and properly narrows the stripped-headers list to match current code. However it needs changes before merge:

**Blocking (factual errors):**
1. Default state: CORS is **enabled**, not disabled (line 6)
2. `corsDefaultAllowedHeaders` default value mismatch (line 47 in configuration-cors.md, line 597 in zowe-yaml-configuration.md)
3. Property name typo `corsDefaultAllowedMethods` should be `corsAllowedMethods` (line 45)

**Important omissions (since this is a "security overhaul"):**
4. PR #4823 (July 2026) added `SecFetchSiteFilter` (token-free CSRF protection via `Sec-Fetch-Site`) and restricted login content-type. Neither is documented.

**Nice-to-haves:**
5. Clarify additive behavior of `corsAllowedOrigins` (custom-metadata.md line 69)
6. Document `Access-Control-Allow-Credentials: true` as hardcoded (configuration-cors.md line 50)
7. Avoid duplicating per-service CORS parameter docs across two pages (onboard-direct-eureka-call.md line 212)

Recommend **Request changes**.

balhar-jakub (dismissed):
Solid documentation overhaul that correctly tracks the metadata keys and narrows the stripped-headers list. However, three factual errors and one significant omission (PR #4823 — Sec-Fetch-Site CSRF protection and login content-type restriction, both shipped after this PR opened) need to be addressed before merge.

See inline comments for specifics. **Requesting changes.**

